PR #31, opened 2023-02-04
[Bug] Fix typecasing problem.

Tang8330:
There's a type casing problem between downstreams apps and JSON.

Imagine the following scenario:
1) `name`
2) `NAME`
3) `NAmE`
4) `NamE`

JSON will treat these as **four** distinct values.
DWH will normalize this (Snowflake will UPPERCASE and BigQuery will LOWERCASE) into `name`, thus 1 value.

Fix here is to remove redundant lowercasing AND be more cognizant around case sensitive operations.


Tang8330:
Postgres and MongoDB both tested.

Also tested for new columns (created by Transfer) and old column combination (inferred from Snowflake)

PR #32, opened 2023-02-06
Bug around column typing

Tang8330:
We were not swapping `tableConfig.Columns()[strings.ToLower(inMemCol)]` with `tableData.InMemoryColumns` if exists.

The data from `tableConfig` is considered SOT.


Tang8330:
Snowflake never had an issue with column typing, but added just in case.

[ ] Snowflake has been tested.
[ ] BigQuery has been tested

Tang8330:
BQ worked. This is being released with 1.4.1

PR #36, opened 2023-02-08
Clean up and remove wal2json limitation

Tang8330:
In this PR, we will be doing the following:

1) Remove `wal2json` as a limitation for PostgreSQL replication slot plug-in. We've tested `pgoutput` and `decoderbufs` in addition to `wal2json`. All of which are supported.

2) Adding additional test coverage around BQ merge


Tang8330:
Additional test coverage added. README also updated to reflect the current state.

Issue #35, opened 2023-02-08
Migrate documentation to Gitbooks


Tang8330:
Search and reading experience is significantly better.

Gitbooks also has CI and previews built-in.

PR #37, opened 2023-02-08
Migrate documentation to gitbooks

Tang8330:
Addresses https://github.com/artie-labs/transfer/issues/35


Tang8330:
Checked all the links and they're working as expected!

PR #38, opened 2023-02-09
Adding .goreleaser

Tang8330:
Integrating with `.goreleaser` so we can have a streamline way to release builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# GoReleaser
+dist/*
+
 *.mock.go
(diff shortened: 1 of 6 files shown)


Tang8330:
- Tested this to generate releases in our Github repo.
- Also generates CHANGELOG
- This also generates docker images.

Tang8330:
Tested the Docker image and the build flag is working as expected to workaround the `m1/m2` problem.

PR #39, opened 2023-02-10
[Enhancement] Improving typing for PostgreSQL

Tang8330:
We were previously only relying on the `JSON` library for parsing Kafka messages to determine typing. As you can see from https://github.com/golang/go/issues/56719

[RFC 8259](https://www.rfc-editor.org/rfc/rfc8259#section-3):
> A JSON value MUST be an object, array, number, or string, or one of the following three literal names:
> * false
> * null
> * true
> There is only number, but not int or float in JSON.

Previous PR added schema support within the message and this PR builds on top of this for PostgreSQL by being able to detect if the specified field is an integer. 

If so, it will typecast within `GetData(...)` and ensure that Transfer's typing library correctly labels the said value as an integer.


Tang8330:
Typing working as expected.
image

image

PR #42, opened 2023-02-16
[Enhancement] Adding the ability to toggle on / off dropping deleted columns in the destination

Tang8330:
In this PR, we are adding the ability for customers to toggle whether they want Transfer to drop deleted columns in the destination.


Tang8330:
Snowflake and BQ both tested

PR #43, opened 2023-02-18
Refactoring DWH Alter Table

Tang8330:
This PR will refactor AlterTable which is implemented by both BigQuery and Snowflake.
The syntax are virtually identical, with the only exceptions being:

1) Typing
2) Errors

This PR is to consolidate the two by leveraging helper functions to distinguish the differences. Doing so will streamline us adding more DWH providers and consolidate testing and minimize code drift.

Last minor thing that got added was that the `snowflake.Store` object now uses `db.Store` as an anonymous struct such that I don't need to write boilerplate for the DWH interface to expose `Exec(...)`


Tang8330:
Code looks good. Not changing functionality. Just cleaning up for subsequent PRs like: https://github.com/artie-labs/transfer/pull/42 and new destinations to be eaasier.

Merging

PR #45, opened 2023-02-19
[Enhancement] Adding the ability to toggle soft deletion

Tang8330:
Right now, Transfer's handling of DML is that it will delete the rows in the destination if the rows have been deleted in the source.

This PR adds an ability to soft delete the row by adding a column: `__artie_delete` (bool).
* If someone turns on soft deletion (off by default), it will add the column above.
* If the row is deleted and soft deletion is turned on, it will set `__artie_delete = true`

### DIFF
--- a/lib/dwh/dml/merge.go
+++ b/lib/dwh/dml/merge.go
@@ -9,7 +9,7 @@ import (
 	"github.com/artie-labs/transfer/lib/config/constants"
 )
 
-func MergeStatement(fqTableName, subQuery, pk, idempotentKey string, cols []string) (string, error) {
+func MergeStatement(fqTableName, subQuery, pk, idempotentKey string, cols []string, softDelete bool) (string, error) {
 	// We should not need idempotency key for DELETE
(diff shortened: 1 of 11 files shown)


lib/dwh/dml/merge.go
```diff
@@ -39,7 +60,7 @@ func MergeStatement(fqTableName, subQuery, pk, idempotentKey string, cols []stri
 	return fmt.Sprintf(`
 			MERGE INTO %s c using (%s) as cc on c.%s = cc.%s
 				when matched AND cc.%s then DELETE
-				when matched AND IFNULL(cc.%s, false) = false %s then UPDATE
+				when matched AND IFNULL(cc.%s, false) = false %sthen UPDATE
```
Tang8330, line 63: This is on purpose. This is to avoid double spacing since the idempotent clause has spacing already.

Tang8330:
Testing against BigQuery and Snowflake working fine.
Tested with both softDelete on and off.

PR #49, opened 2023-02-25
Fix HTML code in string column and double quote in JSON data

TamasNo1:
Opening this pull request to fix some issues I run into while setting up sync between PostgreSQL 12.11 and Snowflake.

Apologies if the tests look weird, I'm not familiar with Golang, so I just followed the pattern I saw in the existing test file.

Basically I saw 2 scenarios when the generated SQL code contained syntax errors:
1. When a text column in the database table contains HTML code that contains escaped javascript code (ie. data got in the DB during penetration testing for XSS attacks)
2. When a JSON column in the database table contains a string value that contains a `"` symbol

### DIFF
--- a/clients/snowflake/merge.go
+++ b/clients/snowflake/merge.go
@@ -14,7 +14,7 @@ import (
 
 func stringWrapping(colVal interface{}) string {
 	// Escape line breaks, JSON_PARSE does not like it.
-	colVal = strings.ReplaceAll(fmt.Sprint(colVal), `\n`, `\\n`)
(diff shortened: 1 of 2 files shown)


Tang8330:
@TamasNo1 Thanks for your contribution! I had to sanitize the line break upon a multi-line text. 

Understood on both your points, XSS and also `"` symbol. Let me run this through my tests again and see if removing it will break the multi-line text. If so, I'll see what we can do about both those use cases you mentioned!

Tang8330:
Code looks great. Thanks for contributing! Tested multiline string, XSS and `"`. All working as expected!

Tang8330 (approved):

PR #50, opened 2023-02-25
Escape strings correctly for BQ

Tang8330:
Follows up from https://github.com/artie-labs/transfer/pull/49, to fix BQ.

Also pulled this into a separate utility function


Tang8330:
Testing multi-line string, XSS and `"` on BQ

Tang8330:
All 3 scenarios working as expected. I'll be sure to add this into our integration test suite that's coming up.

/cc @TamasNo1

PR #54, opened 2023-03-02
Swap where we source table name from

Tang8330:
# Context

Previously, we were fetching the table name from within the message payload itself. Now, we are fetching it directly from the config file.

# Why

We are doing this so that we can support Kafka tombstones (the message will be null, so we will not be able to derive the name from there).

This should be a relatively minor change - however, the in-memory database will now store the table with the table name derived from the destination vs. src.

Scenario #1 - My source table is called `orders` and the table in Snowflake is called `PREFIX_orders`. Our in-memory DB will be saving data against `PREFIX_orders`.

Scenario #2 - Both table names are the same, no changes.

### DIFF
--- a/lib/cdc/event.go
+++ b/lib/cdc/event.go
@@ -14,7 +14,6 @@ type Format interface {
 }
 
 type Event interface {
-	Table() string
 	GetExecutionTime() time.Time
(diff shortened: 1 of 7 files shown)


Tang8330:
BigQuery and Postgres working as expected.

Issue #48, opened 2023-02-21
Filter tombstone messages from Debezium


Tang8330:
Tombstone messages has no data in the body, let's filter it out.

PR #55, opened 2023-03-02
Handling Kafka Tombstones

Tang8330:
In this PR, we are adding support for Kafka tombstones.

This allows users to leverage Kafka tombstone + compaction to optimize performance even further.

Motivation and considerations linked [here](https://docs.google.com/document/d/1ApCs6EewoIlUUHkJGDAJLOs7f13AyB2AhuAAk020FJM/edit?usp=sharing)

### DIFF
--- a/clients/snowflake/snowflake.go
+++ b/clients/snowflake/snowflake.go
@@ -2,7 +2,6 @@ package snowflake
 
 import (
 	"context"
-
 	"github.com/snowflakedb/gosnowflake"
 
 	"github.com/artie-labs/transfer/lib/config"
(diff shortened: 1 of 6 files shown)


Tang8330:
Pre-req PR: https://github.com/artie-labs/transfer/pull/54

Tang8330:
Working as expected.

Tang8330:
Note: This does not impact DDL statements because the execution time will be null, it will not be higher than the watermark expected to drop columns.

Tang8330:
Addresses: https://github.com/artie-labs/transfer/issues/48

PR #61, opened 2023-03-06
Ability to use GCP Pub/Sub instead of Kafka

Tang8330:
This PR enables Artie Transfer to consume from GCP Pub/Sub instead of Kafka.

Pub/Sub has functional equivalency to Kafka and we are following the same pattern (i.e. consume async, then acking when flush is successful).

When this PR is merged, we'll then update our documentation accordingly.

Work remaining:
- [x] Tests
- [x] Update README
- [ ] Examples
- [ ] New Image


Tang8330:
Examples will be a fast follow

Tang8330:
Tests working as expected.

PR #73, opened 2023-04-07
BigQuery <> MongoDB equality fix

Tang8330:
BigQuery cannot compare equality of two objects without having some typecasting to a STRING.

Transfer only does equality comparisons with primary keys which should not be JSON structs with the exception to MongoDB which may use `objectID`.


Tang8330:
Tests and manual QA passed

Issue #68, opened 2023-03-14
Allow a third variable to be specified to trigger flush


Tang8330:
Check the in-memory DB and if the size exceeds X, then flush.

Effectively, the equation becomes, whichever one is sooner:
* Flush interval (10s)
* Numbers of rows within the in-memory DB (15k)
* Size of the in-memory DB

Tang8330:
Is it 1 MB? https://docs.snowflake.com/en/user-guide/query-size-limits

Confirm.

Tang8330:
cc: @TamasNo1 - Have you run into this issue before? 

Thinking about picking this up within the next few days.

TamasNo1:
Hey @Tang8330 I literally fixed this a couple of days ago, still verifying if it works smoothly!
In our case we don't have a large number of rows, but rather large JSON fields in some tables, so even though the row counting was my first idea as well, I decided to check the potential query length and do the limitation that way:

https://github.com/sitestream/transfer/commit/1129c622e8ea8775eedd89c0cb35a004913a0c24

I'm not sure if this is the nicest solution, just wanted to make a quick proof of concept, but so far it seems to be working!

Tang8330:
Ah good to know!

Tang8330:
Done with PR #89

PR #89, opened 2023-05-08
Flush rules to be cognizant of in-memory database size

Tang8330:
Addresses https://github.com/artie-labs/transfer/issues/68

This is necessary because Snowflake and BigQuery both have their limits on how big (in kb) a request can be. Snowflake's API docs specify 1 MB [src](https://docs.snowflake.com/en/user-guide/query-size-limits)...but the actual limit is significantly higher.

Through this PR, I've tested 25 mb payloads without any issue. However, this does happen...especially for tables with TOAST columns. The current workaround is to just reduce the number of items from the `bufferRow` but that's not robust.

This PR will allow users to specify an additional merge rule, and our merge rules will now look like:
1. Flush interval time (default 10s)
2. Pool rows (default to 15k) - number of rows.
3. Pool size (default to 25 mb) - size of the in-memory db.

Whichever one is invoked first will trigger a flush cycle.

TODOs:
- [x] Update documentation -> https://docs.artie.so/running-transfer/options
- [x] Provide benchmarking
- [x] Tests
- [x] Impl

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,3 +28,4 @@ release:
 .PHONY: bench_size
 bench_size:
 	go test ./lib/size -bench=Bench -benchtime=10s
+	go test ./lib/optimization -bench=Bench -benchtime=10s
--- a/Makefile
+++ b/Makefile
@@ -24,3 +24,7 @@ build:
 .PHONY: release
 release:
 	goreleaser release --clean
+
+.PHONY: bench_size
+bench_size:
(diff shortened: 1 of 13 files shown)


Tang8330:
Will confirm this alleviates the REQUEST ENTITY error from Snowflake.

Tang8330:
FYI @TamasNo1 - this is now officially supported.

TamasNo1:
Amazing, well done @Tang8330 🎉

PR #111, opened 2023-05-31
Snowflake to use TIMESTAMP with TIME ZONE instead of TIMESTAMP_NTZ

Tang8330:
Snowflake's `TIMESTAMP_NTZ` (of which datetime is an alias for) - is buggy when it comes to JOINs and working with other datasets that uses TIME ZONE (where TZ = UTC).

As mentioned in the comments, Snowflake will try to parse this column out into the session's local TZ and then parse it back into UTC which messes with the actual values.

### DIFF
--- a/lib/typing/snowflake.go
+++ b/lib/typing/snowflake.go
@@ -1,8 +1,9 @@
 package typing
 
 import (
-	"github.com/artie-labs/transfer/lib/typing/ext"
 	"strings"
+
+	"github.com/artie-labs/transfer/lib/typing/ext"
 )
 
 type SnowflakeKind string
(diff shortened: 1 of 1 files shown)


Tang8330:
image
image

Tang8330:
This is now included in the `2.0.0` version.

For folks that need to migrate the existing columns, follow this command.

```sql
-- 1. Add a new column with TIMESTAMP_TZ data type
ALTER TABLE customers.public.customers_test ADD COLUMN ts_tz_test_new TIMESTAMP_TZ;

-- 2. Update the new column with the converted data
UPDATE customers.public.customers_test SET ts_tz_test_new = CONVERT_TIMEZONE('UTC', CONCAT(ts_tz_test::text, ' +0000'));

-- 3. Drop the old column
ALTER TABLE customers.public.customers_test DROP COLUMN ts_tz_test;

-- 4. Rename the new column to the old column's name
ALTER TABLE customers.public.customers_test RENAME COLUMN ts_tz_test_new TO ts_tz_test;
```

PR #115, opened 2023-06-04
Supporting SMT routing

Tang8330:
[Guide to Debezium's SMT topic routing](https://debezium.io/documentation/reference/stable/transformations/topic-routing.html)

## TL;DR

By default, Debezium is mapped 1:1 for topic to physical table. If you would like to change this - you'll need to modify this with SMT topic routing.

Why would you do this?

* Tables with multiple partitions go to same topic
* Multiple tables going to one topic (especially for Pub/Sub where topics are not automatically created).

## Changes

* This PR was made possible by making `tableName` within the `topicConfig` to be optional from this PR: https://github.com/artie-labs/transfer/pull/113.

* We also added the `projectID` prefix to our BigQuery fully qualified name logic since BigQuery may not play well if the table and topic are not in the same project. 

* Filtering out Debezium Topic Routing Key `__dbz__physicalTableIdentifier` which is used enforce uniqueness within the topic.

### DIFF
--- a/clients/bigquery/ddl.go
+++ b/clients/bigquery/ddl.go
@@ -11,8 +11,8 @@ import (
 	"github.com/artie-labs/transfer/lib/typing"
 )
 
-func (s *Store) getTableConfig(_ context.Context, tableData *optimization.TableData) (*types.DwhTableConfig, error) {
-	fqName := tableData.ToFqName(constants.BigQuery)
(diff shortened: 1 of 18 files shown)


Tang8330:
This is included in `v2.0.3`

https://docs.artie.so/tutorials/debezium-topic-reroute-smt

PR #117, opened 2023-06-08
Escaping column names for DWHs

Tang8330:
## TL;DR

We are now escaping column names that are reserved keywords. We are starting with `start` and `select`. We have created an util function since we only need to escape this when making actual calls to Snowflake and BigQuery.

For this, we created `typing.NameArgs` as BigQuery and Snowflake require different types of escaping.

Snowflake wants it to be like: 
```sql
SELECT "start" from table;
```

BigQuery wants it to be like: 
```sql
SELECT `select` from table;
```

This was sprinkled all over the codebase which explains the large diff. This has been tested against JSON, ARRAY and TOASTED columns.


Tang8330:
Tested this against BigQuery and Snowflake

PR #122, opened 2023-06-13
Supporting Decimal Handling Mode

Tang8330:
# Problem

1. We do not support all [decimal.handling.mode](https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-property-decimal-handling-mode) options. This PR enables us to support them all by adding support for `precise`.
2. `SMALLINT` and `SMALLSERIAL` were getting casted as `FLOAT`
3. We are storing large numbers as `float64` which has support up to around ~`e15-e16` before precision loss occurs. This PRs adds support for calling:

```
NUMERIC(p, s), where:
 * p = precision
 * s = scale
```

# Changes

1. Correctly assessing `SMALLINT` and `SMALLSERIAL`
2. Consolidated `typing/Makefile` into the main make and invoked via `make bench_typing`
3. Changed the query we're making to fetch table configurations on BigQuery to be more robust. Previous way required extensive escaping and parsing logic.

**New SQL query:**
```sql
SELECT column_name, data_type FROM  `project.INFORMATION_SCHEMA.COLUMNS` WHERE table_name="table";
```

4. We're adding the `*decimal.Decimal` data object which is a wrapper around `*big.Float` and stores precision and scale metadata. 

* `.String()` reports the whole data object back with no precision loss.
* `.Value()` will return the value as STRING (these strategies are done to eliminate precision loss):
  * when source precision exceeds the destination precision
  * when source scale exceeds the destination scale 
  * when the precision is not specified by the source (e.g. literal `NUMERIC` and `DECIMAL`) which indicates a variable precision decimal. 
* Else, `.Value()` will return it back in `*big.Float`

# Checks before merge:
- [x] Backwards compatibility test
- [x] Additional test coverage
- [x] Benchmarking on `decimal.Decimal` and parse


Tang8330:
This is the benchmark stats.

```bash
go test ./lib/debezium -bench=Bench -benchtime=20s
goos: darwin
goarch: amd64
pkg: github.com/artie-labs/transfer/lib/debezium
cpu: VirtualApple @ 2.50GHz
BenchmarkDecodeDecimal_P64_S10-12    	11332038	      2152 ns/op
BenchmarkDecodeDecimal_P38_S2-12     	18427053	      1311 ns/op
BenchmarkDecodeDecimal_P5_S2-12      	20784638	      1153 ns/op
```

Tang8330:
It's a bit slower than I would have hoped, happy path would be just around 1/1000th of a ms per column to compute.
However, this is already extremely optimized, further would require the use of memoization, but I'm doubtful that it'll help given the cardinality of this will be extremely high.

PR #124, opened 2023-06-14
Fix panic

Tang8330:
This PR https://github.com/artie-labs/transfer/pull/121 previously had patched a panic. By doing so, it has uncovered another potential panic which this PR is patching.


Tang8330:
Also adding the 2 security patches.

PR #143, opened 2023-07-02
Address dependabot alerts

Tang8330:
* Upgrading [Snowflake driver to 1.6.23](https://github.com/artie-labs/transfer/security/dependabot/3) (after the arrow 12.0.1 upgrade which is a patch for 32bit builds) 
* Upgrade [jwx](https://github.com/artie-labs/transfer/security/dependabot/4)


Tang8330:
Go releaser passed.

PR #171, opened 2023-08-07
Minor improvement on S3

Tang8330:
## Changes

1. Updating README
2. Deprecating `credentialsClause` which is useful in the Redshift context, but not useful for S3.

## Deprecating credentialsClause

Instead, we'll use the traditional `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` and allow them to pass it in. Upon which, we will override our S3 library.


Tang8330:
Code LGTM

PR #207, opened 2023-10-23
[WIP] Kafka commit retries

Tang8330:


Tang8330:
Just deployed this as a canary. Once this works, will merge.

processes/consumer/configs.go
```diff
@@ -44,7 +48,25 @@ func CommitOffset(ctx context.Context, topic string, partitionsToOffset map[stri
...
 			if msg.KafkaMsg != nil {
-				err = topicToConsumer.Get(topic).CommitMessages(ctx, *msg.KafkaMsg)
+				for attempts := 0; attempts < 5; attempts++ {
+					err = topicToConsumer.Get(topic).CommitMessages(ctx, *msg.KafkaMsg)
+					if err == nil {
+						break
+					}
+
+					if kafkalib.IsRetryableErr(err) {
+						jitterDuration := jitter.JitterMs(1500, attempts)
```
nathan-artie, line 58: Could make `jitter.JitterMs` return a `time.Duration`.

processes/consumer/configs.go
```diff
@@ -44,7 +48,25 @@ func CommitOffset(ctx context.Context, topic string, partitionsToOffset map[stri
...
+						logger.FromContext(ctx).WithError(err).WithFields(map[string]interface{}{
+							"attempts":            attempts,
+							"sleep duration (ms)": jitterDuration,
+						}).Info("encountered a retryable error, will sleep and retry...")
+
+						time.Sleep(time.Duration(jitterDuration) * time.Millisecond)
+					} else if err != nil {
```
nathan-artie, line 65: `err` will never be `nil` here because of the check on line 53.

nathan-artie (approved):
LG!, left a few comments.

processes/consumer/configs.go
Tang8330: Yeah, I can make it `else` instead of `else if`

processes/consumer/configs.go
Tang8330: True

nathan-artie (approved):

PR #215, opened 2023-11-08
[WIP] Support downcasting

Tang8330:


Tang8330:
This is not needed because it's actually already supported.

PR #218, opened 2023-11-09
[MongoDB] Escaping infinity and negative infinity correctly

Tang8330:
We were already escaping `NaN`, but we were not escaping `Infinity` and `-Infinity`

This PR adds that capability

### DIFF
--- a/lib/typing/mongo/bson.go
+++ b/lib/typing/mongo/bson.go
@@ -21,9 +21,8 @@ import (
 // JSONEToMap will take JSONE data in bytes, parse all the custom types
 // Then from all the custom types,
 func JSONEToMap(val []byte) (map[string]interface{}, error) {
-	// RegEx on the actual value of `NaN` only (raw value and quotes).
-	// This is because we cannot use RegEx to find only NaN.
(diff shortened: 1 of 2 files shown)


lib/typing/mongo/bson.go
```diff
@@ -21,9 +21,8 @@ import (
 // JSONEToMap will take JSONE data in bytes, parse all the custom types
 // Then from all the custom types,
 func JSONEToMap(val []byte) (map[string]interface{}, error) {
-	// RegEx on the actual value of `NaN` only (raw value and quotes).
-	// This is because we cannot use RegEx to find only NaN.
-	re := regexp.MustCompile(`\bNaN\b|"\bNaN\b"`)
+	// We are escaping `NaN`, `Infinity` and `-Infinity` (literal values)
+	re := regexp.MustCompile(`\bNaN\b|"\bNaN\b"|-\bInfinity\b|"-\bInfinity\b|\bInfinity\b|"\bInfinity\b"`)
```
nathan-artie, line 25: Is there a missing quote here after the second `Infinity\b`?

nathan-artie (approved):

PR #222, opened 2023-11-16
MongoDB edge case

Tang8330:
## Problem

A primary key inside of MongoDB is always `_id` which should be an unique value. If you don't set anything inside of `_id`, MongoDB will fill it with an `ObjectID` [source](https://stackoverflow.com/a/3299005).

* We are relying on the fact that whatever is being written to the Kafka partition key is the primary key of the collection or table. 
* However, Debezium is writing `_id` as `id` in the partition key

This is why in our [code](https://github.com/artie-labs/transfer/blob/daf5b34bc039094f793842a429f663f8438b32f4/lib/cdc/mongo/debezium.go#L142-L146), we have another conditional block that will set whatever is in the Kafka partition key into the event payload to ensure we have uniqueness when save this data.

What this means for our MongoDB syncs is that, we'll have:
```json
{
    "_id": "123",
    "id": "123",
    "foo": "bar"
}
```

Where `_id` and `id` are both present. 

This is problematic is customers also have an `id` field, as we are overwriting it with our guarantee.

## Questions 

> So why don't we get rid of the guarantee is the `id` field is set?

We cannot do this because a MongoDB delete may emit a [tombstone message](https://docs.google.com/document/d/1ApCs6EewoIlUUHkJGDAJLOs7f13AyB2AhuAAk020FJM/edit#heading=h.7vple58ktdjj), which will not contain the message body.

As a result, we will run into consistency issues where for INSERT and UPDATEs, we'll infer whatever is set in `id`, but for DELETEs, we'll issue our own value into `id` which will not match up.

> Can't we just rename `id` to `_id`?

This will break for our existing customers that are running MongoDB because the value that is being parsed from the partition key looks like:

```json
{"$oid":"61990f9568180100242fb76f"}
```

## Solution

1. We need to rename the key from `id` to `_id`
2. We need to correctly parse the data inside of the partition key and run it through our BSON decoder. However, we will need to account for the fact that Kafka's partition key serialization does not handle nested objects.


Tang8330:
Benchmark results
```
Robins-MacBook-Pro:transfer robintang$ make bench_mongo
go test ./lib/cdc/mongo -bench=Bench -benchtime=20s
goos: darwin
goarch: amd64
pkg: github.com/artie-labs/transfer/lib/cdc/mongo
cpu: VirtualApple @ 2.50GHz
BenchmarkGetPrimaryKey-12    	  509916	     45855 ns/op
PASS
ok  	github.com/artie-labs/transfer/lib/cdc/mongo	24.516s
```

TL;DR -- taking `0.045 ms/op`

nathan-artie (approved):

PR #223, opened 2023-11-18
[Redshift] Edge case with schema inference

Tang8330:
## Problem

Say we are syncing from MongoDB and the column can be both JSON and STRING.

In Redshift, we are casting this column as `VARCHAR(MAX)` since `SUPER` doesn't like certain non-ASCII characters. 
Transfer's inference library will use Redshift as the source of truth and cast all these values down into STRING.

However the JSON object that is being casted as a string looks like `map[foo] = bar` as opposed to `{"foo": "bar"}`

### DIFF
--- a/clients/redshift/cast.go
+++ b/clients/redshift/cast.go
@@ -23,7 +23,7 @@ const (
 
 // replaceExceededValues - takes `colVal` interface{} and `colKind` columns.Column and replaces the value with an empty string if it exceeds the max length.
 // This currently only works for STRING and SUPER data types.
-func replaceExceededValues(colVal string, colKind columns.Column) interface{} {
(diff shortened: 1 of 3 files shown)


clients/redshift/cast.go
```diff
@@ -112,7 +124,7 @@ func (s *Store) CastColValStaging(ctx context.Context, colVal interface{}, colKi
 
 	// Checks for DDL overflow needs to be done at the end in case there are any conversions that need to be done.
 	if s.skipLgCols {
-		colValString = fmt.Sprint(replaceExceededValues(colVal, colKind))
+		colValString = fmt.Sprint(replaceExceededValues(colValString, colKind))
```
Tang8330, line 127: This needed to be colValString which has been parsed already

Issue #227, opened 2023-11-22
BigQuery temp tables are not dropped after merge


claytongentry:
v. 2.0.58

Transfer generates BigQuery temp tables with suffixes that include capital letters, e.g. `my_table___artie_dM9YWyeakM`.

However, those suffixes are downcased in the drop statements, e.g. `my_table___artie_dm9ywyeakm`. This has the effect that the temporary table is not dropped at all, as it is does not exist. Lots of temp tables accrete in the dataset until they expire several hours later.

claytongentry:
Looks great, thank you!

PR #228, opened 2023-11-22
[BigQuery bug] Fix dropping temporary tables

Tang8330:
## Problem

Addresses https://github.com/artie-labs/transfer/issues/227

The problem here is that `stringutil.Random(...)` could return capital letters, but our DROP TABLE statement is lowercasing it for Snowflake and Redshift.

This leads to temporary tables not being deleted in BigQuery, also BigQuery does not require us to escape table names, below SQL statement works.

```sql
DROP TABLE foo.bar_artie_AAA;
```


Tang8330:
Edit: Updated PR description to be more clear.

PR #235, opened 2023-12-16
[Redshift] Ability to infer VARCHAR length

Tang8330:
## Problem

Debezium does not expose the character precision within the actual message. For example, we don't know if the column is `VARCHAR(123) vs VARCHAR(333)`.

As such, we have been creating all string based columns as `VARCHAR(MAX)` to preserve precision. However this comes with a tradeoff of over-provisioning space. 

For Snowflake and other data warehouses, it doesn't matter - the actual space used will be the actual length of the value. However, Redshift allocates space based on the column precision.

The workaround for this has been to go into the actual table in DWH and update it yourself:

```sql
ALTER TABLE public.customers_test2 ALTER COLUMN first_name TYPE VARCHAR(123);
```

When this is done, Transfer will not try to mutate the column back to `VARCHAR(MAX)` and will just continue to reference it accordingly. However, when it is still using `VARCHAR(MAX)` to create string-based columns for delta tables.

## Change

* This PR will infer the character precision and pass that context to creation of delta tables.
* Renamed `UpdateInMemoryColumnsFromDestination` to `MergeColumnsFromDestination` to be more clear

### DIFF
--- a/lib/typing/redshift.go
+++ b/lib/typing/redshift.go
@@ -1,19 +1,31 @@
 package typing
 
 import (
+	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/artie-labs/transfer/lib/typing/ext"
 )
 
-func RedshiftTypeToKind(rawType string) KindDetails {
+func RedshiftTypeToKind(rawType string, stringPrecision string) KindDetails {
 	rawType = strings.ToLower(rawType)
 	if strings.HasPrefix(rawType, "numeric") {
 		return ParseNumeric(defaultPrefix, rawType)
 	}
 
+	// TODO: Add test
 	if strings.Contains(rawType, "character varying") {
-		return String
+		var strPrecision *int
+		precision, err := strconv.Atoi(stringPrecision)
+		if err == nil {
+			strPrecision = &precision
+		}
+
+		return KindDetails{
+			Kind:                         String.Kind,
+			OptionalRedshiftStrPrecision: strPrecision,
+		}
 	}
 
 	switch rawType {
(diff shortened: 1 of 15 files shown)


lib/typing/redshift_test.go
```diff
@@ -3,63 +3,101 @@ package typing
...
 	}
 
 	for _, testCase := range testCases {
-		for _, rawType := range testCase.rawTypes {
-			kd := RedshiftTypeToKind(rawType)
+		for _, rawTypeAndPrec := range testCase.rawTypes {
+			kd := RedshiftTypeToKind(rawTypeAndPrec.rawType, rawTypeAndPrec.precision)
 			assert.Equal(t, testCase.expectedKd.Kind, kd.Kind, testCase.name)
+
+			if kd.OptionalRedshiftStrPrecision != nil {
```
nathan-artie, line 98: Can we just do:
`assert.Equal(t, *testCase.expectedKd.OptionalRedshiftStrPrecision, *kd.OptionalRedshiftStrPrecision, testCase.name)`
Without the check that `kd.OptionalRedshiftStrPrecision ` is not `nil`, which will also ensure that we're not failing to set a `kd.OptionalRedshiftStrPrecision` for a test case that expects one.

nathan-artie: Or I suppose add an `else` with `assert.Nil(t, *kd.OptionalRedshiftStrPrecision, testCase.name)`

lib/typing/redshift.go
```diff
@@ -44,11 +55,17 @@ func kindToRedShift(kd KindDetails) string {
 		return "INT8"
 	case Struct.Kind:
 		return "SUPER"
-	case String.Kind, Array.Kind:
+	case Array.Kind:
```
nathan-artie, line 58: Nit: should we call this function `kindToRedshift` with the "S" for "shift" lowercased?

nathan-artie (approved):

lib/typing/redshift.go
Tang8330: Yep, thanks

PR #238, opened 2023-12-22
Optimizing the way we calculate our in-memory pool size

Tang8330:
# TL;DR
giphy

## Problem

Previously, we were encoding the values to be string and then calculating the number of bytes to approximate how large each row is.

For super wide tables, it was pretty slow and could take up to 20ms each time it was invoked.

## Change

Now, we are iterating over the row and then calculating each row by first inspecting the primitive type and if it's indeed a string, just count the number of bytes as opposed to serializing it out again.

## Results

```
BEFORE
cpu: VirtualApple @ 2.50GHz
BenchmarkGetApproxSize_TallTable-12    	    1604	   7524981 ns/op
BenchmarkGetApproxSize_WideTable-12    	     214	  54952219 ns/op

BenchmarkTableData_ApproxSize_TallTable-12    	 7741323	      1654 ns/op
BenchmarkTableData_ApproxSize_WideTable-12    	 1000000	     11223 ns/op

-----
AFTER
BenchmarkGetApproxSize_TallTable-12    	   35970	    333518 ns/op
BenchmarkGetApproxSize_WideTable-12    	     745	  15972217 ns/op

BenchmarkTableData_ApproxSize_TallTable-12    	19568608	       743.2 ns/op
BenchmarkTableData_ApproxSize_WideTable-12    	 2203387	      6018 ns/op
```

* After is 1.8x faster for wide tables (the bottom 2 tests)
* After is 2.2x faster for tall tables

### DIFF
--- a/lib/size/size.go
+++ b/lib/size/size.go
@@ -1,19 +1,13 @@
 package size
 
-import "fmt"
+import (
+	"fmt"
+)
 
-func GetApproxSize(v map[string]interface{}) int {
+func GetApproxSize(value interface{}) int {
 	// We chose not to use unsafe.SizeOf or reflect.Type.Size (both are akin) because they do not do recursive traversal.
 	// We also chose not to use gob.NewEncoder because it does not work for all data types and had a huge computational overhead.
 	// Another plus here is that this will not error out.
-	var size int
-	for _, value := range v {
-		size += approximateSizeOfValue(value)
-	}
-	return size
-}
-
-func approximateSizeOfValue(value interface{}) int {
 	switch v := value.(type) {
 	case string:
 		return len(v)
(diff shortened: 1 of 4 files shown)


lib/size/size_test.go
```diff
@@ -23,12 +19,9 @@ func TestVariableToBytes(t *testing.T) {
...
-
-	stat, err := os.Stat(filePath)
-	assert.NoError(t, err)
-
 	size := GetApproxSize(rowsData)
-	assert.Equal(t, int(stat.Size()), size)
+
+	// Check if size is non-zero and seems plausible
+	assert.NotZero(t, size, "Size should not be zero")
+	assert.Greater(t, size, 1000, "Size should be reasonably large for the given data structure")
```
Tang8330, line 26: Previous test was encoding it into a string and then writing the file.

lib/size/size.go
```diff
@@ -1,14 +1,41 @@
 package size
 
-import (
-	"fmt"
-)
+import "fmt"
 
 // GetApproxSize will encode the actual variable into bytes and then check the length (approx) by using string encoding.
```
nathan-artie, line 5: This isn't accurate anymore.

lib/size/size.go
```diff
@@ -1,14 +1,41 @@
...
+		size += approximateSizeOfValue(value)
+	}
+	return size
+}
+
+func approximateSizeOfValue(value interface{}) int {
+	switch v := value.(type) {
+	case string:
+		return len(v)
+	case []byte:
```
nathan-artie, line 21: Could combine this with the previous case.

nathan-artie (approved):

nathan-artie (approved):

PR #246, opened 2024-01-10
Go 1.21

nathan-artie:


lib/stringutil/strings.go
```diff
@@ -55,6 +54,5 @@ func stringWithCharset(length int, charset string) string {
 }
 
 func Random(length int) string {
-	rand.Seed(time.Now().UnixNano())
```
nathan-artie, line 58: from staticcheck:
> Error: lib/stringutil/strings.go:58:2: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: As of Go 1.20 there is no reason to call Seed with a random value. Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator.  (SA1019)

Tang8330 (approved):

PR #250, opened 2024-01-12
Refactoring TableData ToFqName

Tang8330:
## Changes

* Refactoring `tableData.ToFqName(...)` to accept another `bigQueryProjectID (if exists)`. Doing this as a pre-cursor to the larger body of work to remove `context` from the typing library.
* Change `ddl.AlterTable` to early exit if there are no columns to alter to improve readability

### DIFF
--- a/clients/bigquery/merge.go
+++ b/clients/bigquery/merge.go
@@ -115,10 +115,12 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 	// Check if all the columns exist in BigQuery
 	srcKeysMissing, targetKeysMissing := columns.Diff(ctx, tableData.ReadOnlyInMemoryCols(),
 		tableConfig.Columns(), tableData.TopicConfig.SoftDelete, tableData.TopicConfig.IncludeArtieUpdatedAt)
+
+	tableName := tableData.ToFqName(ctx, s.Label(), true, s.projectID)
 	createAlterTableArgs := ddl.AlterTableArgs{
 		Dwh:         s,
 		Tc:          tableConfig,
-		FqTableName: tableData.ToFqName(ctx, s.Label(), true),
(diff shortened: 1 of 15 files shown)


clients/bigquery/merge.go
```diff
@@ -227,7 +229,7 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 	}
 
 	mergeQuery, err := dml.MergeStatement(ctx, &dml.MergeArgument{
-		FqTableName:               tableData.ToFqName(ctx, constants.BigQuery, true),
+		FqTableName:               fqName,
```
nathan-artie, line 232: In the original code the second argument was `constants.BigQuery` but now it's going to be `s.Label()`.

Tang8330: You're right! I wanted to use `s.Label()` since it does the same thing and it consolidates our usage of `constants.BigQuery`, see code: https://github.com/artie-labs/transfer/blob/abb9ca73258b3c014b404f609fe035a137948f5a/clients/bigquery/bigquery.go#L59

nathan-artie: Ah I hadn't gotten around to looking at the source for `Label()` yet, good call.

clients/snowflake/staging.go
```diff
@@ -115,7 +115,7 @@ func (s *Store) mergeWithStages(ctx context.Context, tableData *optimization.Tab
 		return nil
 	}
 
-	fqName := tableData.ToFqName(ctx, constants.Snowflake, true)
+	fqName := tableData.ToFqName(ctx, s.Label(), true, "")
```
nathan-artie, line 118: I assume this is the same thing as with BigQuery, but just calling it out.

Tang8330: yup!

PR #252, opened 2024-01-15
[Refactor] Adding RawValue functions

Tang8330:
## Changes

* Adding `col.RawDefaultValue()` which will just return the variable without any manipulation (used for logging)
* Adding `col.RawName()` which will also just return the name without any manipulation
* Adding `tableData.RawName()`

By doing so, we are able to remove `context` from intermediary functions and simplify the codepaths. This PR will tee us up to remove `context.Context` from our Typing library.

### DIFF
--- a/clients/bigquery/bigquery.go
+++ b/clients/bigquery/bigquery.go
@@ -38,7 +38,7 @@ func (s *Store) getTableConfig(ctx context.Context, tableData *optimization.Tabl
 		FqName:    tableData.ToFqName(ctx, s.Label(), true),
 		ConfigMap: s.configMap,
 		Query: fmt.Sprintf("SELECT column_name, data_type, description FROM `%s.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS` WHERE table_name='%s';",
-			tableData.TopicConfig.Database, tableData.Name(ctx, nil)),
+			tableData.TopicConfig.Database, tableData.RawName()),
 		ColumnNameLabel:    describeNameCol,
 		ColumnTypeLabel:    describeTypeCol,
 		ColumnDescLabel:    describeCommentCol,
--- a/clients/redshift/merge.go
+++ b/clients/redshift/merge.go
@@ -84,7 +84,7 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 
 		err = utils.BackfillColumn(ctx, s, col, tableData.ToFqName(ctx, s.Label(), true))
 		if err != nil {
-			return fmt.Errorf("failed to backfill col: %v, default value: %v, error: %v", col.RawName(), col.RawDefaultValue(), err)
+			return fmt.Errorf("failed to backfill col: %v, default value: %v, err: %v", col.RawName(), col.RawDefaultValue(), err)
 		}
 
 		tableConfig.Columns().UpsertColumn(col.RawName(), columns.UpsertColumnArg{
--- a/clients/redshift/merge.go
+++ b/clients/redshift/merge.go
@@ -68,7 +68,7 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 	}
 
 	tableConfig.AuditColumnsToDelete(srcKeysMissing)
-	tableData.MergeColumnsFromDestination(ctx, tableConfig.Columns().GetColumns()...)
+	tableData.MergeColumnsFromDestination(tableConfig.Columns().GetColumns()...)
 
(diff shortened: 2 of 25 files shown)


clients/redshift/merge.go
```diff
@@ -84,12 +84,10 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 
 		err = utils.BackfillColumn(ctx, s, col, tableData.ToFqName(ctx, s.Label(), true))
 		if err != nil {
-			defaultVal, _ := col.DefaultValue(ctx, nil)
-			return fmt.Errorf("failed to backfill col: %v, default value: %v, error: %v",
-				col.Name(ctx, nil), defaultVal, err)
+			return fmt.Errorf("failed to backfill col: %v, default value: %v, error: %v", col.RawName(), col.RawDefaultValue(), err)
```
nathan-artie, line 87: `%w` for error?

clients/bigquery/bigquery.go
```diff
@@ -38,7 +38,7 @@ func (s *Store) getTableConfig(ctx context.Context, tableData *optimization.Tabl
 		FqName:    tableData.ToFqName(ctx, s.Label(), true),
 		ConfigMap: s.configMap,
 		Query: fmt.Sprintf("SELECT column_name, data_type, description FROM `%s.INFORMATION_SCHEMA.COLUMN_FIELD_PATHS` WHERE table_name='%s';",
-			tableData.TopicConfig.Database, tableData.Name(ctx, nil)),
+			tableData.TopicConfig.Database, tableData.RawName()),
```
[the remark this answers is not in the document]
Tang8330: Yep, we don't need to escape this

image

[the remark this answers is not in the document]
Tang8330: > Does tableName come directly from the database or does it come from a source that a user could manipulate and add inject SQL into?

It comes from 2 places.

1. Kafka message (produced by Debezium)
2. Table name override set by config.yaml

It is possible to do a SQL injection, I'll add something to our backlog to gather an inventory of potential SQL injections and come up with mitigants

lib/destination/types/table_config.go
```diff
@@ -56,30 +56,30 @@ func (d *DwhTableConfig) Columns() *columns.Columns {
...
+func (d *DwhTableConfig) MutateInMemoryColumns(createTable bool, columnOp constants.ColumnOperation, cols ...columns.Column) {
 	d.Lock()
 	defer d.Unlock()
 	switch columnOp {
 	case constants.Add:
 		for _, col := range cols {
 			d.columns.AddColumn(col)
 			// Delete from the permissions table, if exists.
-			delete(d.columnsToDelete, col.Name(ctx, nil))
+			delete(d.columnsToDelete, col.RawName())
```
nathan-artie, line 67: After a casual check it seems like `d.columnsToDelete` is only used for tests, and is `nil` otherwise. Do we need it?

Tang8330: It's being used by internal functions to provide a read only version for external workloads e.g. `ReadOnlyColumnsToDelete`

clients/redshift/merge.go
Tang8330: Let's keep it the same for now? There's not much benefit from changing this yet

clients/snowflake/staging.go
```diff
@@ -175,12 +175,10 @@ func (s *Store) mergeWithStages(ctx context.Context, tableData *optimization.Tab
 
 		err = utils.BackfillColumn(ctx, s, col, tableData.ToFqName(ctx, s.Label(), true))
 		if err != nil {
-			defaultVal, _ := col.DefaultValue(ctx, nil)
-			return fmt.Errorf("failed to backfill col: %v, default value: %v, error: %v",
-				col.Name(ctx, nil), defaultVal, err)
+			return fmt.Errorf("failed to backfill col: %v, default value: %v, error: %v", col.RawName(), col.RawDefaultValue(), err)
```
nathan-artie, line 178: `err: %w`

clients/redshift/merge.go
nathan-artie: Also `err` instead of `error`, I think that's our convention?

lib/destination/types/table_config.go
nathan-artie: Ohh `AddColumnsToDelete` is mutating it, I had missed that before.

nathan-artie (approved):

clients/redshift/merge.go
Tang8330: True will fix

PR #255, opened 2024-01-17
Sanitizing JSON inputs from RDBMS

Tang8330:


lib/jsonutil/jsonutil.go
```diff
@@ -0,0 +1,24 @@
+package jsonutil
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+func SanitizePayload(val interface{}) (interface{}, error) {
```
nathan-artie, line 8: Add TODO to extend this for `[]interface{}`.

lib/jsonutil/jsonutil.go
```diff
@@ -0,0 +1,24 @@
...
+	_, isOk := val.(map[string]interface{})
+	if isOk {
+		return val, nil
+	}
+
+	var jsonMap map[string]interface{}
+	err := json.Unmarshal([]byte(fmt.Sprint(val)), &jsonMap)
```
nathan-artie, line 15: Can we just cast to a string instead of using `fmt.Sprint()`? We can actually just return `val` if it isn't a string in the first place.

lib/jsonutil/jsonutil_test.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	{
+		// Valid JSON string, nothing changed.
+		val, err := SanitizePayload(`{"hello":"world"}`)
+		assert.NoError(t, err)
+		assert.Equal(t, `{"hello":"world"}`, val)
+	}
+	{
+		// Fake JSON - appears to be in JSON format, but has duplicate keys
+		val, err := SanitizePayload(`{"hello":"world","hello":"world"}`)
```
nathan-artie, line 25: Can we make this first value something other than "world" so that we're testing the behavior that the second key overrides the first?

lib/jsonutil/jsonutil_test.go
```diff
@@ -0,0 +1,52 @@
...
+		for key := range jsonMap {
+			if key == "hello" {
+				foundHello = true
+			}
+			if key == "foo" {
+				foundFoo = true
+			}
+		}
+
+		assert.True(t, foundHello)
```
nathan-artie, line 49: Can you do:
```
assert.Contains(g.T(), jsonMap, "hello")
assert.Contains(g.T(), jsonMap, "foo")
```
?

Tang8330: yup

lib/jsonutil/jsonutil.go
```diff
@@ -0,0 +1,24 @@
+package jsonutil
+
+import (
+	"encoding/json"
+)
+
+// SanitizePayload will take in a JSON string, and return a JSON string that has been sanitized (removed duplicate keys)
+func SanitizePayload(val interface{}) (interface{}, error) {
+	var jsonMap map[string]interface{}
```
nathan-artie, line 9: Can move `var jsonMap map[string]interface{}` to after the `val.(string)` check.

nathan-artie (approved):

PR #259, opened 2024-01-18
[ExtTime] Enforce YYYY - Take two

Tang8330:
## Motivation

Bad data can come into Transfer, we should figure out a way to filter them, especially for `time.Time` objects where the year exceeds the YYYY format or in BC. They will get rejected by downstream destinations, so let's filter them out and not hard fail.

### DIFF
--- a/lib/debezium/types.go
+++ b/lib/debezium/types.go
@@ -60,29 +60,38 @@ func RequiresSpecialTypeCasting(typeLabel string) (bool, SupportedDebeziumType) 
 	return false, Invalid
 }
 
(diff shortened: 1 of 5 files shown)


lib/debezium/types.go
```diff
@@ -60,29 +60,43 @@ func RequiresSpecialTypeCasting(typeLabel string) (bool, SupportedDebeziumType)
...
+		extTime, err = ext.NewExtendedTime(time.UnixMicro(val).In(time.UTC), ext.TimeKindType, "")
+	default:
+		return nil, fmt.Errorf("supportedType: %s, val: %v failed to be matched", supportedType, val)
+	}
 
+	if err != nil {
+		return nil, err
+	}
+
+	if extTime != nil {
```
nathan-artie, line 93: Could combine if statements.

PR #265, opened 2024-01-19
Remove `context` from `lib/typing` - Part 4

Tang8330:
As per title, this is the fourth (should be one more after) to remove `context `from the Typing library. This will bring more stability to external services that want to reference Transfer's typing library programmatically without hitting run-time panics.

## Changes
* `uppercaseEscName` is extracted when we load the data warehouse and injected as a private variable on the DWH object
* Refactored some merge functions to make it more ergonomic

## After
The last PR will be to remove `context` from being used in `ParseValue`

### DIFF
--- a/clients/s3/s3.go
+++ b/clients/s3/s3.go
@@ -28,7 +28,8 @@ import (
 )
 
 type Store struct {
-	Settings *config.S3Settings
+	Settings          *config.S3Settings
+	uppercaseEscNames bool
 }
 
 func (s *Store) Validate() error {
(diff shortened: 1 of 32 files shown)


clients/s3/s3.go
```diff
@@ -146,9 +147,10 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 	return os.RemoveAll(fp)
 }
 
-func LoadStore(ctx context.Context, settings *config.S3Settings) (*Store, error) {
+func LoadStore(_ context.Context, settings *config.S3Settings) (*Store, error) {
```
nathan-artie, line 150: Can context be removed as an arg or does it need to be here to be compatible with an interface?

clients/snowflake/snowflake.go
```diff
@@ -97,17 +98,21 @@ func (s *Store) reestablishConnection(ctx context.Context) {
...
 	if _store != nil {
 		// Used for tests.
 		return &Store{
-			testDB:    true,
-			Store:     *_store,
-			configMap: &types.DwhToTablesConfigMap{},
+			testDB:            true,
+			uppercaseEscNames: cfg.SharedDestinationConfig.UppercaseEscapedNames,
+			configMap:         &types.DwhToTablesConfigMap{},
+			Store:             *_store,
```
nathan-artie, line 109: Do you want to add a line of whitespace here between the private and public variables like in the other store initializers?

clients/snowflake/snowflake_test.go
```diff
@@ -110,7 +110,7 @@ func (s *SnowflakeTestSuite) TestExecuteMergeReestablishAuth() {
 		tableData.InsertRow(pk, row, false)
 	}
 
-	s.stageStore.configMap.AddTableToConfig(tableData.ToFqName(s.ctx, constants.Snowflake, true, ""),
+	s.stageStore.configMap.AddTableToConfig(tableData.ToFqName(constants.Snowflake, true, s.stageStore.uppercaseEscNames, ""),
```
nathan-artie, line 113: `s.stageStore.Label()` instead of `constants.Snowflake`

nathan-artie (approved):

clients/s3/s3.go
Tang8330: Not required!

PR #272, opened 2024-01-22
Pass typing settings to ParseValue

nathan-artie:


lib/typing/typing.go
```diff
@@ -12,6 +11,10 @@ import (
 	"github.com/artie-labs/transfer/lib/typing/ext"
 )
 
+type TypingSettings struct {
```
nathan-artie, line 14: @Tang8330 Is this what you were talking about this morning?

lib/cdc/postgres/debezium.go
```diff
@@ -1,19 +1,19 @@
...
 	"github.com/artie-labs/transfer/lib/config/constants"
 	"github.com/artie-labs/transfer/lib/debezium"
 	"github.com/artie-labs/transfer/lib/kafkalib"
+	"github.com/artie-labs/transfer/lib/typing"
 )
 
 type Debezium string
 
-func (d *Debezium) GetEventFromBytes(ctx context.Context, bytes []byte) (cdc.Event, error) {
+func (d *Debezium) GetEventFromBytes(typingSettings typing.Settings, bytes []byte) (cdc.Event, error) {
```
Tang8330, line 16: Since this variable isn't used, you can do `GetEventFromBytes(_ typing.Settings, bytes []byte)`

lib/config/config.go
```diff
@@ -94,12 +95,7 @@ type SharedDestinationConfig struct {
 }
 
 type SharedTransferConfig struct {
-	AdditionalDateFormats []string `yaml:"additionalDateFormats"`
-
-	// CreateAllColumnsIfAvailable - If true, we will create all columns if the metadata is available regardless of
-	// whether we have a value from the column. This will also bypass our Typing library.
-	// This only works for data sources with a schema such as Postgres and MySQL
-	CreateAllColumnsIfAvailable bool `yaml:"createAllColumnsIfAvailable"`
+	typing.Settings
```
Tang8330, line 98: Woohoo!

Tang8330: @nathan-artie Can you add a test to make sure the yaml gets unmarshalled correctly w/ the anonymous struct?

Tang8330 (approved):

lib/config/config.go
nathan-artie: Hah they don't, going to add `ToTypingSettings` function instead.

PR #273, opened 2024-01-22
[PostGIS] Supporting Geometry

Tang8330:
In addition to supporting `Point`, we should be able to support `Geometry`.

image

### DIFF
--- a/lib/cdc/util/geometry_test.go
+++ b/lib/cdc/util/geometry_test.go
@@ -1,6 +1,8 @@
 package util
 
-import "github.com/stretchr/testify/assert"
+import (
+	"github.com/stretchr/testify/assert"
+)
 
 func (u *UtilTestSuite) TestParseGeometryPoint() {
 	{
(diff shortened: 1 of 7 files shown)


lib/cdc/util/geometry_test.go
```diff
@@ -15,3 +17,15 @@ func (u *UtilTestSuite) TestParseGeometryPoint() {
...
+
+func (u *UtilTestSuite) TestGeometryWkb() {
+	{
+		geoJSONString, err := parseGeometry(map[string]interface{}{
+			"wkb":  "AQEAAAAAAAAAAADwPwAAAAAAAPA/",
+			"srid": nil,
+		})
+
+		assert.NoError(u.T(), err)
+		assert.Equal(u.T(), `{"type":"Feature","geometry":{"type":"Point","coordinates":[1,1]},"properties":null}`, geoJSONString)
```
Tang8330, line 29: null exists here because the geojson library we're using to parse wkb doesn't have `omitempty`

nathan-artie (approved):

PR #285, opened 2024-01-29
Include database updated at + Improving topic config validation

Tang8330:
## Motivation

In addition to including `__artie_updated_at`, we will now also emit `__artie_database_updated_at` to give you the exact timestamp from the database on which this event was processed.

## Changes

* Including a new table config option
* Setting the value if enabled
* Topic Config validation to return a descriptive error as opposed to `bool` 

## Note

In order to not cause issues, we'll need to deploy this PR first: https://github.com/artie-labs/transfer/pull/286


lib/typing/columns/diff_test.go
```diff
@@ -49,10 +50,22 @@ func (c *ColumnsTestSuite) TestShouldSkipColumn() {
...
 			softDelete:            true,
 		},
+		{
+			name:                     "db updated at col BUT setting is not enabled",
+			colName:                  constants.DatabaseUpdatedColumnMarker,
+			includeDatabaseUpdatedAt: false,
+			expectedResult:           true,
+		},
+		{
+			name:                     "db updated at col AND setting is not enabled",
```
nathan-artie, line 60: Nit: `setting is enabled`.

lib/cdc/util/relational_event.go
```diff
@@ -81,24 +81,25 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]interface{}, tc *kafkalib.
...
-		}
-
 		for k, v := range pkMap {
 			retMap[k] = v
 		}
 
 		// If idempotency key is an empty string, don't put it in the payload data
 		if tc.IdempotentKey != "" {
-			retMap[tc.IdempotentKey] = s.GetExecutionTime().Format(time.RFC3339)
+			retMap[tc.IdempotentKey] = s.GetExecutionTime().Format(ext.ISO8601)
```
Tang8330, line 90: Be consistent

lib/kafkalib/topic.go
```diff
@@ -65,20 +66,24 @@ func (t *TopicConfig) String() string {
 		t.Database, t.Schema, t.TableName, t.Topic, t.IdempotentKey, t.CDCFormat, t.DropDeletedColumns)
 }
 
-func (t *TopicConfig) Valid() bool {
+func (t *TopicConfig) Validate() error {
```
Tang8330, line 69: Returning an error instead of a boolean so that we can be more descriptive when it fails to validate.

lib/kafkalib/topic.go
```diff
@@ -65,20 +66,24 @@ func (t *TopicConfig) String() string {
...
 
 	// IdempotentKey is optional.
 	empty := array.Empty([]string{t.Database, t.Schema, t.Topic, t.CDCFormat})
 	if empty {
-		return false
+		return fmt.Errorf("database, schema, topic or cdc format is empty")
 	}
 
 	if t.CDCKeyFormat == "" {
 		t.CDCKeyFormat = defaultKeyFormat
```
nathan-artie, line 81: This is a little gross, the `Validate` method shouldn't be mutating the topic config. Maybe add a method for this as a FF?

lib/kafkalib/topic_test.go
```diff
@@ -117,7 +117,7 @@ func TestTopicConfig_String(t *testing.T) {
 
 func TestTopicConfig_Validate(t *testing.T) {
 	var tc TopicConfig
-	assert.False(t, tc.Valid(), tc.String())
+	assert.Error(t, tc.Validate(), tc.String())
```
nathan-artie, line 120: Use `assert.ErrorContains` with the expected messages?

Tang8330: :+1:

lib/kafkalib/topic.go
Tang8330: Agreed, chatted offline. Will do a FF

nathan-artie (approved):

PR #286, opened 2024-01-29
Skipping tombstone messages

Tang8330:
## Context

Depending on how Debezium is configured - it may emit a Kafka tombstone message upon each delete. Previously Artie Transfer was treating it as a delete event. This is redundant and could cause issues when we want to start emitting the database updated timestamp (since the Kafka message will be `nil`).

## Changes

Delete a row only when the database says to delete the row. Skip tombstone messages altogether.

### DIFF
--- a/processes/consumer/kafka.go
+++ b/processes/consumer/kafka.go
@@ -111,24 +111,18 @@ func StartConsumer(ctx context.Context) {
 			topicToConsumer.Add(topic, kafkaConsumer)
 			for {
 				kafkaMsg, err := kafkaConsumer.FetchMessage(ctx)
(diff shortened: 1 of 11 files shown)


processes/consumer/kafka.go
```diff
@@ -124,6 +124,11 @@ func StartConsumer(ctx context.Context) {
 					continue
 				}
 
+				if len(msg.Value()) == 0 {
+					slog.With(logFields...).Info("found a tombstone message, skipping...")
```
nathan-artie, line 128: I think you can do `slog.Info("found a tombstone message, skipping...", logFields...))`, the other ones use `With` because we're sending along an error too.

Tang8330: cool

nathan-artie (approved):

PR #289, opened 2024-01-30
Pull context up in a couple places

nathan-artie:


clients/redshift/merge.go
```diff
@@ -83,7 +84,7 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 			continue
 		}
 
-		err = utils.BackfillColumn(ctx, s, col, fqName)
+		err = utils.BackfillColumn(*config.FromContext(ctx).Config, s, col, fqName)
 		if err != nil {
```
Tang8330, line 88: Can we inject the Store object with `config` and just pass it around when necessary?

nathan-artie: You mean pass store to `utils.BackfillColumn`?

nathan-artie: And also `TableData`?

Tang8330: Sorry for not being clear.

I meant have `config` as an object on store, such that this function becomes something like:

```go
err = utils.BackfillColumn(s.Config, s, col, fqName)
```

nathan-artie: Yep, will do that in a follow up.

PR #295, opened 2024-01-31
Add config to BigQuery store

nathan-artie:


clients/bigquery/bigquery.go
```diff
@@ -32,6 +32,7 @@ type Store struct {
 	batchSize         int
 	projectID         string
 	uppercaseEscNames bool
```
Tang8330, line 34: Don't need project ID or uppercaseEscName?

nathan-artie: True, I'll remove.

Tang8330 (approved):

PR #301, opened 2024-01-31
[config] Removing unnecessary pointers

Tang8330:
The config object which is nested under `Settings` had a pointer. This PR removes it.

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -118,7 +118,7 @@ func (k *Kafka) String() string {
 		k.BootstrapServer, k.GroupID, k.Username != "", k.Password != "")
 }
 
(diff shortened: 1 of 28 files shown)


lib/config/config.go
```diff
@@ -118,7 +118,7 @@ func (k *Kafka) String() string {
 		k.BootstrapServer, k.GroupID, k.Username != "", k.Password != "")
 }
 
-func (c *Config) TopicConfigs() ([]*kafkalib.TopicConfig, error) {
+func (c Config) TopicConfigs() ([]*kafkalib.TopicConfig, error) {
```
Tang8330, line 121: Also removed all the pointers from these functions since it's not actually mutating the object

nathan-artie (approved):

PR #308, opened 2024-01-31
[Topic Config] Validation Fast Follow

Tang8330:
## Changes

Transfer now requires you to explicitly set `cdcKeyFormat`, whereas before our validation would change the format to a `KeyFormatString` if the value is empty.


lib/debezium/keys.go
```diff
@@ -5,24 +5,23 @@ import (
...
 
+	"github.com/artie-labs/transfer/lib/kafkalib"
+
 	"github.com/artie-labs/transfer/lib/typing/columns"
 
 	"github.com/artie-labs/transfer/lib/config/constants"
 )
 
 const (
-	KeyFormatJSON   = "org.apache.kafka.connect.json.JsonConverter"
```
Tang8330, line 14: Extras, just removed em

nathan-artie (approved):

PR #310, opened 2024-02-01
Skipping Operations Improvement

Tang8330:
## Changes

This PR adds the ability to skip operations other than deletes, such as:
* `r` - replication / backfill
* `u` - update
* `c` - create
* `d` - delete

This can be done by specifying at the `topicConfig` level:

```yaml
kafka:
  bootstrapServer: localhost:29092
  groupID: transfer4
  username: ""
  password: ""
  enableAWSMKSIAM: false
  topicConfigs:
    - db: customers
      tableName: users_test2
      schema: public
      skippedOperations: c, r, d
```

This also adds a backwards compatibility port from `skipDelete`.

We now have feature parity with Debezium's [skipped.operations](https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-property-skipped-operations)

### DIFF
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -362,8 +362,8 @@ kafka:
  username: %s
  password: %s
  topicConfigs:
-  - { db: customer, tableName: orders, schema: public}
-  - { db: customer, tableName: customer, schema: public}
+  - { db: customer, tableName: orders, schema: public, skippedOperations: d}
+  - { db: customer, tableName: customer, schema: public, skippedOperations: c}
 
 snowflake:
  account: %s
@@ -398,24 +398,28 @@ reporting:
 	assert.Equal(t, password, config.Kafka.Password)
 	assert.True(t, config.SharedTransferConfig.TypingSettings.CreateAllColumnsIfAvailable)
 
-	var foundOrder bool
-	var foundCustomer bool
+	orderIdx := -1
+	customerIdx := -1
+	for idx, topicConfig := range config.Kafka.TopicConfigs {
+		topicConfig.Load()
 
-	for _, topicConfig := range config.Kafka.TopicConfigs {
 		assert.Equal(t, topicConfig.Database, "customer")
 		assert.Equal(t, topicConfig.Schema, "public")
 
 		if topicConfig.TableName == "orders" {
-			foundOrder = true
+			orderIdx = idx
 		}
 
 		if topicConfig.TableName == "customer" {
-			foundCustomer = true
+			customerIdx = idx
 		}
 	}
 
-	assert.True(t, foundCustomer)
-	assert.True(t, foundOrder)
+	assert.True(t, config.Kafka.TopicConfigs[orderIdx].ShouldSkip("d"))
+	assert.True(t, config.Kafka.TopicConfigs[customerIdx].ShouldSkip("c"))
+
+	assert.True(t, customerIdx >= 0)
+	assert.True(t, orderIdx >= 0)
 
 	// Verify Snowflake config
 	assert.Equal(t, snowflakeUser, config.Snowflake.Username)
--- a/lib/config/config_test.go
+++ b/lib/config/config_test.go
@@ -134,14 +134,15 @@ bufferRows: 10
 	assert.Equal(t, config.FlushIntervalSeconds, 15)
 	assert.Equal(t, int(config.BufferRows), 10)
 
-	assert.Nil(t, config.Validate())
-
 	tcs, err := config.TopicConfigs()
 	assert.NoError(t, err)
 	assert.Equal(t, 2, len(tcs))
 	for _, tc := range tcs {
+		tc.Load()
 		assert.Equal(t, "customer", tc.Database)
 	}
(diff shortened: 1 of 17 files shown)


lib/kafkalib/topic.go
```diff
@@ -56,6 +63,34 @@ const (
...
+	// Operations that we support today:
+	// 1. c - create
+	// 2. r - replication (backfill)
+	// 3. u - update
+	// 4. d - delete
+
+	t.opsToSkipMap = make(map[string]bool)
+	skippedOps := strings.Split(t.SkippedOperations, ",")
+	for _, op := range skippedOps {
+		t.opsToSkipMap[strings.TrimSpace(op)] = true
```
nathan-artie, line 76: Could also lowercase the op as well.

lib/kafkalib/topic.go
```diff
@@ -56,6 +63,34 @@ const (
...
+	}
+
+	// TODO: For backwards compatibility, remove in a later version.
+	if t.SkipDelete {
+		t.opsToSkipMap["d"] = true
```
nathan-artie, line 81: Could add a warn here.

lib/kafkalib/topic.go
```diff
@@ -56,6 +63,34 @@ const (
...
+	}
+}
+
+func (t TopicConfig) ShouldSkip(op string) bool {
+	if t.opsToSkipMap == nil {
+		slog.Warn("opsToSkipMap is nil, calling Load()")
```
nathan-artie, line 87: Is this message out of date?

nathan-artie: It'd be safer to panic here.

Tang8330: i like the panic

lib/config/config_test.go
```diff
@@ -391,24 +398,28 @@ reporting:
...
+			customerIdx = idx
 		}
 	}
 
-	assert.True(t, foundCustomer)
-	assert.True(t, foundOrder)
+	assert.True(t, config.Kafka.TopicConfigs[orderIdx].ShouldSkip("d"))
+	assert.True(t, config.Kafka.TopicConfigs[customerIdx].ShouldSkip("c"))
+
+	assert.True(t, customerIdx >= 0)
```
nathan-artie, line 421: I feel like this check should happen before `assert.True(t, config.Kafka.TopicConfigs[customerIdx].ShouldSkip("c"))`

Tang8330: Shure

Tang8330: It'll all get evaluated, we're not using require which would fatal on the first failed assertion

lib/kafkalib/topic_test.go
```diff
@@ -138,3 +141,35 @@ func TestTopicConfig_Validate(t *testing.T) {
...
+
+func TestTopicConfig_Load_ShouldSkip(t *testing.T) {
+	{
+		// Test backwards compat
+		tc := TopicConfig{
+			SkipDelete: true,
+		}
+
+		tc.Load()
+		assert.True(t, tc.ShouldSkip("d"), tc.String())
```
nathan-artie, line 153: Should also probably test that passing "d" in SkippedOperations works.

lib/config/config_test.go
nathan-artie: Ohh that's right, carry on!

lib/optimization/event_test.go
```diff
@@ -267,6 +267,27 @@ func TestTableData_ShouldFlushRowLength(t *testing.T) {
...
 
+func TestTableData_ContainsHardDeletes(t *testing.T) {
+	{
+		// Hard delete = true
+		td := NewTableData(nil, nil, kafkalib.TopicConfig{}, "foo")
+		assert.Equal(t, 0, int(td.Rows()))
+
+		td.InsertRow("123", nil, true)
+		assert.Equal(t, 1, int(td.Rows()))
+		assert.True(t, td.containsHardDeletes)
```
nathan-artie, line 278: Call `ContainsHardDeletes()` since we have a method for it?

nathan-artie: Could also test inserting something that isn't a delete.

lib/optimization/event_test.go
```diff
@@ -267,6 +267,27 @@ func TestTableData_ShouldFlushRowLength(t *testing.T) {
...
+	}
+	{
+		// TopicConfig has soft delete turned on, so hard delete = false
+		td := NewTableData(nil, nil, kafkalib.TopicConfig{SoftDelete: true}, "foo")
+		assert.Equal(t, 0, int(td.Rows()))
+
+		td.InsertRow("123", nil, true)
+		assert.Equal(t, 1, int(td.Rows()))
```
nathan-artie, line 286: Can do `assert.Len()`.

nathan-artie (approved):

lib/optimization/event_test.go
Tang8330: sg

PR #312, opened 2024-02-03
Adding History Mode

Tang8330:
## Overview

Running Artie Transfer in History mode means that there will be a separate table called: `table__history` that will record every single data mutation that has happened.

If you are running Artie Transfer in Production, you can run History mode by using a separate Kafka consumer group. 

For more details about History Mode, check out our [blog post](https://www.artie.so/blogs/history-table).

## What does it look like in Snowflake?

image

## Changes required:

- [x] Append table name `__history`
- [x] Implement Append Only
- [x] Configuration validation
- [ ] Documentation (will follow up in a separate PR)
- [x] Adding a new column for `op`
- [x] Tests


clients/redshift/append.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/artie-labs/transfer/lib/optimization"
+)
+
+func (s *Store) Append(ctx context.Context, tableData *optimization.TableData) error {
+	return fmt.Errorf("redshift did not implement this yet")
```
nathan-artie, line 11: Add ":" after redshift or remove from https://github.com/artie-labs/transfer/pull/312/files#diff-1de29085b7c8e7e13085ab7bfae3cc2f58f0dd73218eae775522730c51e8d465R11

lib/config/config.go
```diff
@@ -203,6 +215,14 @@ func readFileToConfig(pathToConfig string) (*Config, error) {
 		config.FlushSizeKb = defaultFlushSizeKb
 	}
 
+	if config.Mode == "" {
```
nathan-artie, line 218: This statement is duplicated.

Tang8330: Thx

nathan-artie: Could also throw an error if the mode is invalid (i.e. not "history" or "replication").

Tang8330: I think doing that for a later version would make more sense

lib/config/config_test.go
```diff
@@ -562,11 +562,23 @@ func TestConfig_Validate(t *testing.T) {
...
+	cfg.Mode = History
+	pubsub.TopicConfigs[0].DropDeletedColumns = true
+	assert.ErrorContains(t, cfg.Validate(), "config is invalid, drop deleted columns is not supported in history mode")
+
+	pubsub.TopicConfigs[0].DropDeletedColumns = false
+	pubsub.TopicConfigs[0].IncludeDatabaseUpdatedAt = false
+	assert.ErrorContains(t, cfg.Validate(), "config is invalid, include database updated at is required in history mode")
+
+	pubsub.TopicConfigs[0].IncludeDatabaseUpdatedAt = true
+	assert.Nil(t, cfg.Validate())
```
nathan-artie, line 575: `assert.NoError`

lib/typing/columns/diff.go
```diff
@@ -23,12 +26,16 @@ func shouldSkipColumn(colName string, softDelete bool, includeArtieUpdatedAt boo
 		return false
 	}
 
+	if mode == config.History && colName == constants.OperationColumnMarker {
```
nathan-artie, line 29: Nit: The other three checks compare `colName` first, swap comparison for consistency.

clients/snowflake/append.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/artie-labs/transfer/lib/destination/ddl"
+	"github.com/artie-labs/transfer/lib/typing/columns"
+
+	"github.com/artie-labs/transfer/lib/optimization"
+)
+
+func (s *Store) Append(ctx context.Context, tableData *optimization.TableData) error {
+	err := s.append(tableData)
+	if IsAuthExpiredError(err) {
+		slog.Warn("authentication has expired, will reload the Snowflake store and retry appending", slog.Any("err", err))
```
nathan-artie, line 17: Capitalize `authentication`

clients/snowflake/append.go
```diff
@@ -0,0 +1,60 @@
...
+		s.reestablishConnection()
+		return s.Append(ctx, tableData)
```
nathan-artie, line 19: In the case that the auth is permanently expired this will recurse forever, should we add some type of max retries?

nathan-artie (approved):

clients/snowflake/append.go
Tang8330: We can add a retry, I'll leave a TODO. We need that for `Merge` too.

Tang8330: We'll need to find a way to keep track of retry count since this function sig is used as part of the interface

nathan-artie (approved):

PR #314, opened 2024-02-04
Return `Duration` from `Jitter()`

nathan-artie:
Also remove hardcoded `maxMs` and require callers to pass one in. This way we can reuse this function in [reader.](https://github.com/artie-labs/reader/)

### DIFF
--- a/clients/bigquery/merge.go
+++ b/clients/bigquery/merge.go
@@ -190,7 +190,7 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 			if TableUpdateQuotaErr(err) {
 				err = nil
 				attempts += 1
-				time.Sleep(jitter.Jitter(1500, 3500, attempts))
+				time.Sleep(jitter.Jitter(1500, jitter.DefaultMaxMs, attempts))
 			} else {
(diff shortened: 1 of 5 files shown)


clients/bigquery/merge.go
```diff
@@ -190,7 +190,7 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 			if TableUpdateQuotaErr(err) {
 				err = nil
 				attempts += 1
-				time.Sleep(time.Duration(jitter.JitterMs(1500, attempts)) * time.Millisecond)
+				time.Sleep(jitter.Jitter(1500, 3500, attempts))
```
Tang8330, line 193: Make 3.5k a constant and use it across the board?

lib/destination/types/table_config_test.go
```diff
@@ -115,7 +114,7 @@ func (t *TypesTestSuite) TestDwhTableConfig_ReadOnlyColumnsToDelete() {
 	for i := 0; i < 100; i++ {
 		wg.Add(1)
 		go func() {
-			time.Sleep(time.Duration(jitter.JitterMs(50, 1)) * time.Millisecond)
+			time.Sleep(time.Duration(rand.Intn(100)) * time.Millisecond)
```
nathan-artie, line 117: Don't need to use Jitter here, these two are equivalent.

Tang8330 (approved):

PR #319, opened 2024-02-07
Dropping staging files and temporary tables regardless of outcome

Tang8330:
# Overview

Today, we are only dropping tables and staging files upon a successful merge. This leaves temporary files and tables lingering around. 

Fortunately for temporary tables, we have a `Sweep(...)` process that will garbage collect old tables. 

## Changes

This PR implements a `defer` such that all temporary tables and files are removed regardless of outcome. We'll still need `Sweep`, but this will reduce our reliance on it.

### DIFF
--- a/clients/bigquery/merge.go
+++ b/clients/bigquery/merge.go
@@ -210,6 +210,13 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 		return fmt.Errorf("failed to insert into temp table: %s, err: %v", tableName, err)
 	}
 
+	defer func() {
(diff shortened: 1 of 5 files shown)


clients/bigquery/merge.go
```diff
@@ -245,9 +252,6 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 	}
 
 	_, err = s.Exec(mergeQuery)
-	// This is above, in the case we have a head of line blocking because of an error
```
nathan-artie, line 248: Can just return `err` here now.

nathan-artie (approved):

PR #321, opened 2024-02-07
Pull metrics client out of context

nathan-artie:


processes/consumer/process.go
```diff
@@ -35,7 +35,7 @@ func processMessage(ctx context.Context, cfg config.Config, inMemDB *models.Data
 	}
 	st := time.Now()
 	defer func() {
```
nathan-artie, line 37: Do we need this `func(){}` here?

Tang8330: Are you just asking about the func wrapper?

As opposed to just doing:
```go

defer metricsClient.Timing("process.message", time.Since(st), tags)
```

nathan-artie: Yep.

Tang8330 (approved):

processes/consumer/process.go
nathan-artie: Turns out you do need it or `time.Since(st)` evaluates to ~0.

PR #324, opened 2024-02-08
Change log level to error for merge failures

nathan-artie:


processes/consumer/flush.go
```diff
@@ -90,7 +90,7 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
 
 			if err != nil {
 				tags["what"] = "merge_fail"
-				slog.With(logFields...).Warn(fmt.Sprintf("Failed to execute %s...not going to flush memory", action), slog.Any("err", err))
+				slog.Error(fmt.Sprintf("Failed to execute %s...not going to flush memory", action), append(logFields, slog.Any("err", err))...)
```
Tang8330, line 93: Does append work?

Tang8330: Also, any reason not to just change `.Warn` to `.Error`? 

```go
slog.With(logFields...).Error(fmt.Sprintf("Failed to execute %s...not going to flush memory", action), slog.Any("err", err))
```

nathan-artie: It should https://go.dev/play/p/yeI7nHhZ9X8?v=goprev

nathan-artie: I can just do the simple level change for now.

Tang8330 (approved):

processes/consumer/flush.go
Tang8330: Good to know

PR #325, opened 2024-02-08
[Redshift] Remove copy statement from error logs.

Tang8330:
Logging copy will leak `credentialsClause`, so let's not. 

Also, changing `fmt.Errorf(...)` to error wrap


clients/redshift/staging.go
```diff
@@ -60,15 +60,15 @@ func (s *Store) prepareTempTable(ctx context.Context, tableData *optimization.Ta
 	})
 
 	if err != nil {
-		return fmt.Errorf("failed to upload this to s3, err: %v", err)
+		return fmt.Errorf("failed to upload this to s3, err: %w", err)
```
nathan-artie, line 63: Nit: could be more specific about what "this" is

Tang8330: True

PR #327, opened 2024-02-09
Add guard that table exists when flushing a single table

nathan-artie:


lib/optimization/table_data.go
```diff
@@ -183,12 +184,7 @@ func (t *TableData) Rows() []map[string]interface{} {
 
 // RowsData will create a read-only map of `rowsData`, this should be strictly used for testing only.
 func (t *TableData) RowsData() map[string]map[string]interface{} {
-	_rowsData := make(map[string]map[string]interface{}, len(t.rowsData))
-	for k, v := range t.rowsData {
-		_rowsData[k] = v
-	}
-
-	return _rowsData
+	return maps.Clone(t.rowsData)
```
nathan-artie, line 187: Just optimized this.

processes/consumer/flush.go
```diff
@@ -32,13 +32,20 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
...
 
-	var wg sync.WaitGroup
 	// Read lock to examine the map of tables
 	inMemDB.RLock()
 	allTables := inMemDB.TableData()
 	inMemDB.RUnlock()
 
-	// Flush will take everything in memory and call Snowflake to create temp tables.
+	if args.SpecificTable != "" {
+		if _, ok := allTables[args.SpecificTable]; !ok {
```
Tang8330, line 41: Why do we need this?

nathan-artie: We're making making sure that we only flush tables that have the same name as `args.SpecificTable`, and that the data for said table isn't empty, but we don't have any checks that the table name that was passed in is actually in the in-memory database. This would only happen if we were to introduce a bug somewhere and pass in a bad value in for `args.SpecificTable`, so this is more of an illegal argument check than it is a check to prevent us from flushing the wrong thing since a bad value for `args.SpecificTable` is a no-op.

Tang8330 (approved):

PR #329, opened 2024-02-09
Tag flush errors with `retryable`

nathan-artie:
So that we can set up different monitor alerts thresholds for `retryable=true` and `retryable=false`.


processes/consumer/flush.go
```diff
@@ -90,6 +90,7 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
 
 			if err != nil {
 				tags["what"] = "merge_fail"
+				tags["retryable"] = fmt.Sprintf("%t", dest.IsRetryableError(err))
```
Tang8330, line 93: Do fmt.Sprint(dest.IsRetryableError(err)) so you don't need to worry about string interpolation?

Tang8330 (approved):

processes/consumer/flush.go
nathan-artie: Good call.

PR #330, opened 2024-02-09
Wrap errors

nathan-artie:


clients/bigquery/merge.go
```diff
@@ -68,7 +68,7 @@ func (s *Store) backfillColumn(column columns.Column, fqTableName string) error
 
 	defaultVal, err := column.DefaultValue(&columns.DefaultValueArgs{Escape: true, DestKind: s.Label()}, additionalDateFmts)
 	if err != nil {
-		return fmt.Errorf("failed to escape default value, err: %v", err)
```
nathan-artie, line 71: I decided to drop `, err` from the error strings when we're logging just the `error`.

Tang8330: It's included above though?

Tang8330 (approved):

clients/bigquery/merge.go
nathan-artie: Yeah above we're logging both `col` and `err` so I left it for now.

PR #340, opened 2024-02-14
Sleep after a merge failure

Tang8330:
## Motivation

Right now, if a merge fails - we will repeatedly hammer the destination to merge. Sometimes this has caused our destination to go down temporarily due to the DDOS.

This function does not have a concept of attempts, let's add a simple sleep in between merge attempts so we don't DDOS downstream destinations.

### DIFF
--- a/processes/consumer/flush.go
+++ b/processes/consumer/flush.go
@@ -98,7 +98,8 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
 			if err != nil {
 				tags["what"] = "merge_fail"
 				tags["retryable"] = fmt.Sprint(dest.IsRetryableError(err))
-				slog.With(logFields...).Error(fmt.Sprintf("Failed to execute %s...not going to flush memory", action), slog.Any("err", err))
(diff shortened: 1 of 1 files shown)


nathan-artie (approved):

processes/consumer/flush.go
```diff
@@ -98,7 +98,8 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
 			if err != nil {
 				tags["what"] = "merge_fail"
 				tags["retryable"] = fmt.Sprint(dest.IsRetryableError(err))
-				slog.With(logFields...).Error(fmt.Sprintf("Failed to execute %s...not going to flush memory", action), slog.Any("err", err))
+				slog.With(logFields...).Error(fmt.Sprintf("Failed to execute %s, not going to flush memory, will sleep for 3 seconds before continuing...", action), slog.Any("err", err))
+				time.Sleep(3 * time.Second)
```
nathan-artie, line 102: We only need to sleep if `dest.IsRetryableError(err)`.

PR #342, opened 2024-02-14
[Redshift] Supporting Append only

Tang8330:
## Motivation

This will enable History Mode for Redshift Append

## Strategy

1. Apply necessary DDLs to the target table
2. Create staging table
3. Run `ALTER TABLE target APPEND staging` https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_TABLE_APPEND.html

### DIFF
--- a/clients/snowflake/append.go
+++ b/clients/snowflake/append.go
@@ -55,6 +55,8 @@ func (s *Store) append(tableData *optimization.TableData) error {
 		return err
 	}
 
+	tableData.MergeColumnsFromDestination(tableConfig.Columns().GetColumns()...)
+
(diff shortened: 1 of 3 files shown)


clients/snowflake/append.go
```diff
@@ -55,6 +55,8 @@ func (s *Store) append(tableData *optimization.TableData) error {
 		return err
 	}
 
+	tableData.MergeColumnsFromDestination(tableConfig.Columns().GetColumns()...)
```
Tang8330, line 58: Missed this line before, this will enable us to downcast columns

nathan-artie (approved):

PR #346, opened 2024-02-16
Debezium Parse Field - Address Arrays

Tang8330:
It already worked, adding tests and removing the TODO.

### DIFF
--- a/lib/cdc/util/parse_test.go
+++ b/lib/cdc/util/parse_test.go
@@ -136,8 +136,16 @@ func TestParseField(t *testing.T) {
 			field: debezium.Field{
 				DebeziumType: string(debezium.JSON),
 			},
-			value:         `[1, 2, 3]`,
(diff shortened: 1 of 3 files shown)


lib/cdc/util/parse_test.go
```diff
@@ -123,6 +123,22 @@ func TestParseField(t *testing.T) {
...
 		},
+		{
+			name: "json",
+			field: debezium.Field{
+				DebeziumType: string(debezium.JSON),
+			},
+			value:         `{"foo": "bar", "foo": "bar"}`,
+			expectedValue: `{"foo":"bar"}`,
+		},
+		{
```
nathan-artie, line 134: IIRC the issue was that we weren't handling arrays of objects or arrays of arrays, can we add a test for those?

nathan-artie (approved):

PR #353, opened 2024-02-21
Add `WithRetries` functions

nathan-artie:


lib/jitter/sleep.go
```diff
@@ -11,6 +12,68 @@ func Jitter(baseMs, maxMs, attempts int) time.Duration {
 	// https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
 	// sleep = random_between(0, min(cap, base * 2 ** attempt))
 	// 2 ** x == 1 << x
+	if maxMs <= 0 {
```
nathan-artie, line 15: `rand.Intn()` panics if the argument you pass it is `0`.

lib/retry/retry.go
```diff
@@ -0,0 +1,84 @@
...
+		if err == nil {
+			return nil
+		} else if !r.isRetryableErr(err) {
+			break
+		}
+	}
+	return err
+}
+
+func WithRetries[T any](retryCfg RetryConfig, f func(attempt int, err error) (T, error)) (T, error) {
```
nathan-artie, line 71: Note this version returns a result.

lib/retry/retry.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type NewRetryConfigArgs struct {
+	JitterBaseMs   int
+	JitterMaxMs    int
+	MaxAttempts    int
+	IsRetryableErr func(err error) bool
+}
+
+func NewRetryConfig(args NewRetryConfigArgs) RetryConfig {
```
Tang8330, line 24: Have the function do some validation and return an error?

(1) Like you mentioned above: `rand.Intn()` will panic if value is less than zero, let's validate that jitter base and max have to be greater than zero.

(2) They have to set `IsRetryableErr` or else every function will just be inefficient since everything will be retried

(3) If `maxAttempts = 0`, this would never run, so we should also check for that.

nathan-artie: The hope was that this would be a reusable function that we could use in `reader` and other repos that could handle different types of retry logic we might want to use for different situations.

1. Allowing jitter max + base to be 0 allows for situations where we might want to retry something but not have sleeps in-between. 
2. There are some places we are currently retrying everything that I was going to use this for, such as postgres scanning in `reader`, and I'd like it to be a drop in replacement. Do you think it's better to require `IsRetryableErr` and pass in a `return true` function in those cases?
3. The `max(args.MaxAttempts, 1)` below takes care of this currently, though I can check for that.

Tang8330: > Allowing jitter max + base to be 0 allows for situations where we might want to retry something but not have sleeps in-between.

Agreed. Do you think that's common though? Or more of an edge case? IMO it's the latter and you can then just specify a very small wait, such that you don't DDOS the underlying f(x).

> There are some places we are currently retrying everything that I was going to use this for, such as postgres scanning in reader, and I'd like it to be a drop in replacement. Do you think it's better to require IsRetryableErr and pass in a return true function in those cases?

Yes, and I would imagine the function would start off as:

```go
IsRetryableErr(err error) bool {
 slog.Info("Retrying on this error...", slog.Any("err", err))
return true
}
```

And then over time we would build up a catalog of retryable errors vs. deterministic errors.

3. Agreed, it also does take care of not having negative numbers, although I think having it more explicit as part of the validation would be more clear to others.

nathan-artie: We wouldn't need to `slog.Info("Retrying on this error...", slog.Any("err", err))` though as we already log every time in `sleepIfNecessary` if we're going to retry.

PR #355, opened 2024-02-23
Supporting SQL Server as a destination

Tang8330:
## Changes

This PR enables Transfer to write to SQL Server as a destination.

## Work left

- [x] Documentation (will be a fast follow)
- [x] Configuration validation
- [x] Sweep
- [x] Comment (out of scope)
- [x] Append (History mode)
- [x] Testing every column type
- [x] Test schema evolution
- [x] Table idempotency


clients/mssql/queries.go
```diff
@@ -0,0 +1,34 @@
...
+        WHEN DATA_TYPE = 'numeric' THEN
+		'numeric(' + COALESCE(CAST(NUMERIC_PRECISION AS VARCHAR), '') + ',' + COALESCE(CAST(NUMERIC_SCALE AS VARCHAR), '') + ')'
+		ELSE
+		DATA_TYPE
+	END AS DATA_TYPE,
+    CHARACTER_MAXIMUM_LENGTH
+FROM 
+    INFORMATION_SCHEMA.COLUMNS
+WHERE 
+    LOWER(TABLE_NAME) = LOWER('%s') AND LOWER(TABLE_SCHEMA) = LOWER('%s');`, rawTableName, schema)
```
nathan-artie, line 23: Can we escape the table name and schema in case they have quotes in them?

clients/mssql/staging.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (s *Store) PrepareTemporaryTable(tableData *optimization.TableData, tableConfig *types.DwhTableConfig, tempTableName string, additionalSettings types.AdditionalSettings) error {
+	tempAlterTableArgs := ddl.AlterTableArgs{
+		Dwh:               s,
+		Tc:                tableConfig,
+		FqTableName:       tempTableName,
+		CreateTable:       true,
+		TemporaryTable:    true,
+		ColumnOp:          constants.Add,
+		UppercaseEscNames: &s.config.SharedDestinationConfig.UppercaseEscapedNames,
```
nathan-artie, line 22: Is it possible to not use `UppercaseEscNames` for MSSQL?

clients/mssql/staging.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	if err := ddl.AlterTable(tempAlterTableArgs, tableData.ReadOnlyInMemoryCols().GetColumns()...); err != nil {
+		return fmt.Errorf("failed to create temp table: %w", err)
+	}
+
+	tx, err := s.Begin()
```
nathan-artie, line 29: Do we need to defer a transaction rollback in the case an error happened?

clients/mssql/staging.go
```diff
@@ -0,0 +1,69 @@
...
+			castedValue, castErr := parseValue(value[col], colKind, additionalDateFmts)
+			if castErr != nil {
+				return castErr
+			}
+
+			row = append(row, castedValue)
+		}
+
+		if _, err = stmt.Exec(row...); err != nil {
+			return fmt.Errorf("failed to copy row, err: %w", err)
```
nathan-artie, line 56: Nit: Remove ", err"

clients/mssql/staging.go
Tang8330: Definitely possible, but I don't think the trade off of reducing migration scope is worth the code fragmentation

lib/config/mssql.go
```diff
@@ -0,0 +1,40 @@
...
+type MSSQL struct {
+	Host     string `yaml:"host"`
+	Port     int    `yaml:"port"`
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
+	Database string `yaml:"database"`
+}
+
+func (m *MSSQL) DSN() string {
+	return fmt.Sprintf("sqlserver://%s:%s@%s:%d?database=%s", m.Username, m.Password, m.Host, m.Port, m.Database)
```
nathan-artie, line 19: Does database need to be query string encoded?

clients/mssql/store.go
```diff
@@ -0,0 +1,91 @@
...
+		ColumnNameLabel:    describeNameCol,
+		ColumnTypeLabel:    describeTypeCol,
+		ColumnDescLabel:    describeDescriptionCol,
+		EmptyCommentValue:  ptr.ToString("<nil>"),
+		DropDeletedColumns: tableData.TopicConfig.DropDeletedColumns,
+	})
+}
+
+func LoadStore(cfg config.Config, _store *db.Store) *Store {
+	if _store != nil {
```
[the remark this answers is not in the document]
Tang8330: Done for MSSQL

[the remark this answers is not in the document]
Tang8330: Pulled this out into a separate PR https://github.com/artie-labs/transfer/pull/357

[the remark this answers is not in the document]
Tang8330: We can't move this to tests because shared pkgs like the `ddl` library are loading Redshift and Snowflake directly.

By removing `store` as a 2nd arg, tests are directly going into `db.Open(...)` which will ping the connection and panic if failure occurs.

lib/config/mssql.go
Tang8330: I believe it does need to be encoded, I changed the DSN format.

clients/mssql/queries.go
Tang8330: Param query is very hard because it's sql driver specific and this is in a shared function

Tang8330: So we'll want all the functions to use it together.

clients/mssql/staging.go
```diff
@@ -0,0 +1,80 @@
...
+
+	defer stmt.Close()
+
+	additionalDateFmts := s.config.SharedTransferConfig.TypingSettings.AdditionalDateFormats
+	for _, value := range tableData.Rows() {
+		var row []any
+		for _, col := range columns {
+			colKind, _ := tableData.ReadOnlyInMemoryCols().GetColumn(col)
+			castedValue, castErr := parseValue(value[col], colKind, additionalDateFmts)
+			if castErr != nil {
```
nathan-artie, line 59: If a `castErr` happens we won't roll back since it's a different variable. We could either change it to err or have a boolean called something like `committed` that we set when commit succeds.

Tang8330: I like `committed` as a var so it's less of a gotcha

clients/shared/table_config.go
```diff
@@ -46,7 +48,16 @@ func GetTableConfig(args GetTableCfgArgs) (*types.DwhTableConfig, error) {
 		return tableConfig, nil
 	}
 
-	rows, err := args.Dwh.Query(args.Query)
+	var rows *sql.Rows
+	var err error
+
+	// TODO: Get everyone to pass in args.
+	if len(args.Args) > 0 {
```
nathan-artie, line 55: I think these two branches may be equivalent.

nathan-artie: [[]string{}](https://go.dev/play/p/UFkPiQN6ONj)

lib/config/mssql.go
```diff
@@ -0,0 +1,51 @@
...
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
+	Database string `yaml:"database"`
+}
+
+func (m MSSQL) DSN() string {
+	query := url.Values{}
+	query.Add("database", m.Database)
+
+	u := &url.URL{
```
nathan-artie, line 23: Nice.

lib/destination/ddl/errors_test.go
```diff
@@ -29,6 +29,18 @@ func (d *DDLTestSuite) TestColumnAlreadyExistErr() {
...
+			name:           "MSSQL, table already exist error",
+			err:            fmt.Errorf(`There is already an object named 'customers' in the database.`),
+			kind:           constants.MSSQL,
+			expectedResult: true,
+		},
+		{
+			name:           "MSSQL, column already exists error",
+			err:            fmt.Errorf("Column names in each table must be unique. Column name 'first_name' in table 'users' is specified more than once."),
+			kind:           constants.MSSQL,
+			expectedResult: true,
```
nathan-artie, line 42: Add a test for a MSSQL error that doesn't return `true`?

lib/destination/ddl/expiry.go
```diff
@@ -32,3 +33,23 @@ func ShouldDelete(comment string) (shouldDelete bool) {
...
+	}
+
+	return shouldDeleteUnix(nameParts[len(nameParts)-1])
+}
+
+func shouldDeleteUnix(unixString string) bool {
+	// TODO: Migrate everyone to use shouldDeleteUnix so we don't need to parse comments.
+	unix, err := strconv.Atoi(unixString)
+	if err != nil {
+		return false
```
nathan-artie, line 50: Should we `slog.Warn` here?

lib/destination/dml/merge.go
```diff
@@ -307,3 +307,91 @@ func (m *MergeArgument) GetStatement() (string, error) {
 			Prefix:    "cc.",
 		})), nil
 }
+
+func (m *MergeArgument) GetMSSQLStatement() (string, error) {
```
nathan-artie, line 311: Add some tests for this method?

lib/destination/dml/merge.go
```diff
@@ -307,3 +307,91 @@ func (m *MergeArgument) GetStatement() (string, error) {
...
+
+	cols := m.ColumnsToTypes.GetColumnsToUpdate(*m.UppercaseEscNames, &sql.NameArgs{
+		Escape:   true,
+		DestKind: m.DestKind,
+	})
+
+	if m.SoftDelete {
+		return fmt.Sprintf(`
+			MERGE INTO %s c using %s as cc on %s
+				when matched %sthen UPDATE
```
nathan-artie, line 336: Nit: uppercase `using`, `when`, and `then`.

lib/destination/dml/merge.go
```diff
@@ -307,3 +307,91 @@ func (m *MergeArgument) GetStatement() (string, error) {
...
+				Separator: ",",
+				Prefix:    "cc.",
+			})), nil
+	}
+
+	// We also need to remove __artie flags since it does not exist in the destination table
+	var removed bool
+	for idx, col := range cols {
+		if col == constants.DeleteColumnMarker {
+			cols = append(cols[:idx], cols[idx+1:]...)
```
nathan-artie, line 362: Will be able to use `slices.Delete` when we upgrade to 1.22.

clients/shared/table_config.go
Tang8330: Great!

lib/destination/dml/merge.go
Tang8330: Adding TODO for a FF

PR #357, opened 2024-02-26
Reducing `store` from LoadStore(...)

Tang8330:
We don't need to specify `store` as a 2nd parameter for `LoadStore(...)`.

* Tests can directly inject a `Store` dependency, which is what this PR did.
* Snowflake still needs it as we support `test` as a valid destination

### DIFF
--- a/lib/destination/utils/load.go
+++ b/lib/destination/utils/load.go
@@ -55,9 +55,9 @@ func DataWarehouse(cfg config.Config, store *db.Store) destination.DataWarehouse
 		}
 		return s
 	case constants.BigQuery:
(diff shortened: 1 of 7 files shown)


lib/destination/utils/load.go
```diff
@@ -55,9 +55,9 @@ func DataWarehouse(cfg config.Config, store *db.Store) destination.DataWarehouse
 		}
 		return s
 	case constants.BigQuery:
-		return bigquery.LoadBigQuery(cfg, store)
```
nathan-artie, line 58: Can we explicitly pass in `nil` for  `store` on line 52 where we do `snowflake.LoadSnowflake(cfg, store)`?

PR #359, opened 2024-02-27
Describing tables with parameterized queries

Tang8330:
## Changes

Starting to move away from string interpolation for SQL queries, which are susceptible to SQL injections and moving towards parameterized queries.

Changes are being made to:
- [x] Redshift
- [x] Snowflake (parameterized query only applies to the WHERE clause, so we're skipping Snowflake)
- [x] BigQuery


clients/redshift/queries.go
```diff
@@ -38,6 +33,6 @@ LEFT JOIN
 LEFT JOIN 
     pg_catalog.pg_description d ON d.objsubid=c.ordinal_position AND d.objoid=c1.oid 
 WHERE 
-    LOWER(c.table_name) = LOWER('%s') AND LOWER(c.table_schema) = LOWER('%s');
-`, constants.StrPrecisionCol, args.RawTableName, args.Schema), nil
+    LOWER(c.table_name) = LOWER($1) AND LOWER(c.table_schema) = LOWER($2);
```
Tang8330, line 36: Still requires $1 because seems like pq is still involved :\

nathan-artie (approved):
🎉

PR #361, opened 2024-02-27
Consolidating Sweep functionality

Tang8330:
## Context

With https://github.com/artie-labs/transfer/pull/360, we are now creating temporary tables with `unix ts` as a suffix. This PR is now building on top of it by consolidating the sweeping logic and removing the reliance on parsing comments which was to maintain backwards compatibility.

## Changes

* Changing the shared sweep function to retrieve the schema name from the query as opposed to from the `dbAndSchemaPair`.

^ We need to do this to support MSSQL as `dbo` is an alias for `public`

* Stood up a shared function for Sweep


clients/redshift/redshift.go
```diff
@@ -41,13 +42,13 @@ func (s *Store) Label() constants.DestinationKind {
...
 }
 
-const (
-	describeNameCol        = "column_name"
-	describeTypeCol        = "data_type"
-	describeDescriptionCol = "description"
-)
-
 func (s *Store) GetTableConfig(tableData *optimization.TableData) (*types.DwhTableConfig, error) {
+	const (
```
Tang8330, line 46: Just moved this inline since these constants are function specific.

clients/shared/sweep.go
```diff
@@ -0,0 +1,42 @@
...
+
+		for rows != nil && rows.Next() {
+			var tableSchema, tableName string
+			if err != nil {
+				err = rows.Scan(&tableSchema, &tableName)
+				return err
+			}
+
+			if ddl.ShouldDeleteFromName(tableName) {
+				err = ddl.DropTemporaryTable(dwh, fmt.Sprintf("%s.%s.%s", dbAndSchemaPair.Database, tableSchema, tableName), true)
```
Tang8330, line 33: Note: Using `tableSchema` from the query result as opposed to `dbAndSchemaPair.Schema`

clients/snowflake/snowflake.go
```diff
@@ -51,6 +52,25 @@ func (s *Store) GetTableConfig(tableData *optimization.TableData) (*types.DwhTab
...
+	}
+
+	queryFunc := func(dbAndSchemaPair kafkalib.DatabaseSchemaPair) (string, []any) {
+		return fmt.Sprintf(`
+SELECT 
+    table_schema, table_name
+FROM
+    %s.information_schema.tables
+WHERE
+    table_schema = UPPER(?) AND table_name ILIKE ?`, dbAndSchemaPair.Database), []any{dbAndSchemaPair.Schema, "%" + constants.ArtiePrefix + "%"}
```
nathan-artie, line 68: Is it ever possible to have a `table_schema` that is not upper case? Should we do `UPPER(table_schema)`?

nathan-artie (approved):

clients/snowflake/snowflake.go
Tang8330: it is if the schema is escaped, but also no harm in adding UPPER(table_schema)

Tang8330:
This is currently on hold, we will merge this later this week

PR #364, opened 2024-02-27
[MSSQL] Increasing the resiliency of INTs and FLOATs

Tang8330:
## Changes

1. Adding tests to `parseValue`
2. Ability to cast colVal back into a number if the destination is a number and value provided is in string.
3. Changing everything `parseValue` from `interface{}` -> `any`

### DIFF
--- a/clients/mssql/values_test.go
+++ b/clients/mssql/values_test.go
@@ -1,0 +1,74 @@
+package mssql
+
+import (
+	"testing"
+
+	"github.com/artie-labs/transfer/lib/typing"
+
+	"github.com/artie-labs/transfer/lib/typing/columns"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseValue(t *testing.T) {
+	{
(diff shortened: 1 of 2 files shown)


clients/mssql/values_test.go
```diff
@@ -0,0 +1,74 @@
...
+		assert.True(t, val.(bool))
+
+		val, err = parseValue(false, columns.NewColumn("bool", typing.Boolean), nil)
+		assert.NoError(t, err)
+		assert.False(t, val.(bool))
+
+		// Should be able to handle string booleans
+		val, err = parseValue("true", columns.NewColumn("bool", typing.Boolean), nil)
+		assert.NoError(t, err)
+		assert.True(t, val.(bool))
```
nathan-artie, line 68: Could use `assert.IsType(t, bool, val)`.

nathan-artie: Do you also want to check if the value was parsed correctly?

nathan-artie (approved):

PR #354, opened 2024-02-23
interface{} -> any

nathan-artie:

PR #366, opened 2024-02-28
Renaming `interface{}` to `any` Part 1

Tang8330:
* Taking over this PR: https://github.com/artie-labs/transfer/pull/354
* `any` is an alias of `interface{}`

### DIFF
--- a/lib/cdc/util/geometry.go
+++ b/lib/cdc/util/geometry.go
@@ -28,7 +28,7 @@ type GeometricShapes string
 
 const Point GeometricShapes = "Point"
 
-// ParseGeometry takes in a map[string]interface{} and returns a GeoJSON string.
+// ParseGeometry takes in a map[string]any and returns a GeoJSON string.
 // This function does not use WKB or SRID and leverages X, Y.
 // https://debezium.io/documentation/reference/stable/connectors/postgresql.html#:~:text=io.debezium.data.geometry.Point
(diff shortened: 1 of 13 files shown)


lib/cdc/util/geometry.go
```diff
@@ -31,10 +31,10 @@ const Point GeometricShapes = "Point"
 // ParseGeometry takes in a map[string]interface{} and returns a GeoJSON string.
```
nathan-artie, line 31: This comment can be updated.

lib/cdc/util/parse.go
```diff
@@ -11,8 +11,8 @@ import (
 	"github.com/artie-labs/transfer/lib/debezium"
 )
 
-// ParseField returns a `parsedValue` as type interface{}
-func parseField(field debezium.Field, value interface{}) interface{} {
+// ParseField returns a `parsedValue` as type any
+func parseField(field debezium.Field, value any) interface{} {
```
nathan-artie, line 15: Do you want to change the one at the end too?

lib/cdc/util/geometry.go
Tang8330: cool

lib/cdc/util/parse.go
Tang8330: sure

nathan-artie (approved):

PR #343, opened 2024-02-15
Consolidating MERGE

Tang8330:
## Changes

Every destination has a `MERGE` function, they all follow the same flow, so let's consolidate it.

Functionality should all be the same.

## Checks prior to merge

- [x] Snowflake
- [x] BigQuery
- [x] Redshift

PR #375, opened 2024-03-08
Redshift ignoring table deletes

Tang8330:
## Changes

Our merge consolidation a few weeks ago accidentally missed `containsHardDeletes` flag from the Redshift merge: https://github.com/artie-labs/transfer/pull/343/files#diff-87c1374c50373010ae6b2a9951ae46abee85163c403b39b0f1ac32d8bac54c61

This was not caught because the previous flag was a type `bool`. I changed it to a pointer so that we can validate that it has been set.

### DIFF
--- a/lib/destination/dml/merge.go
+++ b/lib/destination/dml/merge.go
@@ -41,7 +41,7 @@ type MergeArgument struct {
 	SoftDelete bool
 	// ContainsHardDeletes is only used for Redshift and MergeStatementParts,
 	// where we do not issue a DELETE statement if there are no hard deletes in the batch
-	ContainsHardDeletes bool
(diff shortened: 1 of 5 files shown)


lib/destination/dml/merge.go
```diff
@@ -79,6 +79,11 @@ func (m *MergeArgument) GetParts() ([]string, error) {
 		return nil, err
 	}
 
+	// ContainsHardDeletes is only used for Redshift, so we'll validate it now
+	if m.ContainsHardDeletes == nil {
```
nathan-artie, line 83: I would move this after the check that this is redshift.

nathan-artie (approved):

PR #379, opened 2024-03-11
Improving Columns Update Query

Tang8330:
## Changes

1. Adding Microsoft SQL Server tests
2. Renaming `ColumnsUpdateQuery` -> `UpdateQuery` since the pkg name is `columns`
3. Breaking out `struct` and normal TOAST column handling into separate functions

### DIFF
--- a/lib/typing/columns/columns.go
+++ b/lib/typing/columns/columns.go
@@ -246,6 +246,7 @@ func (c *Columns) DeleteColumn(name string) {
 // bigQueryTypeCasting - We'll need to escape the column comparison if the column's a struct.
 // It then returns a list of strings like: cc.first_name=c.first_name,cc.last_name=c.last_name,cc.email=c.email
 func UpdateQuery(columns []string, columnsToTypes Columns, destKind constants.DestinationKind, uppercaseEscNames bool) string {
+	// TODO: We most likely don't need to pass in columns since we already have it from `Columns`.
 	columnsToTypes.EscapeName(uppercaseEscNames, &sql.NameArgs{
 		Escape:   true,
 		DestKind: destKind,
--- a/lib/typing/columns/columns.go
+++ b/lib/typing/columns/columns.go
@@ -240,12 +240,12 @@ func (c *Columns) DeleteColumn(name string) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


lib/typing/columns/columns.go
```diff
@@ -310,3 +271,32 @@ func ColumnsUpdateQuery(columns []string, columnsToTypes Columns, destKind const
 
 	return strings.Join(_columns, ",")
 }
+
+func processToastStructCol(colName string, destKind constants.DestinationKind) string {
+	if destKind == constants.BigQuery {
```
nathan-artie, line 276: Could use switch/case.

lib/typing/columns/columns.go
```diff
@@ -310,3 +271,32 @@ func ColumnsUpdateQuery(columns []string, columnsToTypes Columns, destKind const
...
+			colName, colName, constants.ToastUnavailableValuePlaceholder,
+			colName, colName)
+	} else if destKind == constants.Redshift {
+		return fmt.Sprintf(`%s= CASE WHEN COALESCE(cc.%s != JSON_PARSE('{"key":"%s"}'), true) THEN cc.%s ELSE c.%s END`,
+			colName, colName, constants.ToastUnavailableValuePlaceholder, colName, colName)
+	} else if destKind == constants.MSSQL {
+		// Microsoft SQL Server doesn't allow boolean expressions to be in the COALESCE statement.
+		return fmt.Sprintf("%s= CASE WHEN COALESCE(cc.%s, {}) != {'key': '%s'} THEN cc.%s ELSE c.%s END",
+			colName, colName, constants.ToastUnavailableValuePlaceholder, colName, colName)
+	} else {
```
nathan-artie, line 287: Is this case just Snowflake? If so it might be better explicitly check that it's Snowflake and then error if it isn't so that when we add a new DWH type we remember to implement this for it.

nathan-artie (approved):

lib/typing/columns/columns.go
Tang8330: It is, I'll leave a TODO, erroring out will require changing the function signature

PR #381, opened 2024-03-12
Refactoring Columns Update Query.

Tang8330:
## Changes

Refactoring this further by taking in less params and adding more tests.


lib/typing/columns/columns_test.go
```diff
@@ -366,10 +366,10 @@ func TestColumns_Mutation(t *testing.T) {
 func TestColumnsUpdateQuery(t *testing.T) {
 	type testCase struct {
 		name           string
-		columns        []string
```
nathan-artie, line 369: Rename `columnsToTypes` to `columns`?

nathan-artie (approved):

lib/typing/columns/columns_test.go
Tang8330: sgtm.

PR #384, opened 2024-03-20
[BigQuery] Escape Project ID for fully qualified name

Tang8330:
We need to escape the project ID since it could contain special characters.

### DIFF
--- a/lib/optimization/table_data.go
+++ b/lib/optimization/table_data.go
@@ -208,7 +208,7 @@ func (t *TableData) ToFqName(kind constants.DestinationKind, escape bool, upperc
 		}))
 	case constants.BigQuery:
 		// The fully qualified name for BigQuery is: project_id.dataset.tableName.
-		return fmt.Sprintf("%s.%s.%s", opts.BigQueryProjectID, t.TopicConfig.Database, t.Name(uppercaseEscNames, &sql.NameArgs{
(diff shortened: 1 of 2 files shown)


lib/optimization/table_data.go
```diff
@@ -208,7 +208,7 @@ func (t *TableData) ToFqName(kind constants.DestinationKind, escape bool, upperc
 		}))
 	case constants.BigQuery:
 		// The fully qualified name for BigQuery is: project_id.dataset.tableName.
-		return fmt.Sprintf("%s.%s.%s", opts.BigQueryProjectID, t.TopicConfig.Database, t.Name(uppercaseEscNames, &sql.NameArgs{
+		return fmt.Sprintf("`%s`.%s.%s", opts.BigQueryProjectID, t.TopicConfig.Database, t.Name(uppercaseEscNames, &sql.NameArgs{
```
nathan-artie, line 211: Should we escape `t.TopicConfig.Database` too?

Tang8330: Yeah we should

Tang8330: This is wrong lol: https://github.com/googleapis/google-cloud-go/blob/1ef5b1917bb9a1271c3fb152413ec0e74163164d/bigquery/table.go#L684C27-L684C29

It is what we originally had

nathan-artie: Hah yeah I wondered about that.

PR #385, opened 2024-03-23
Improved data handling of deleted rows

Tang8330:
## Context

Previously, we discarded the `before` object and only wrote the bare minimum into the row such as:

* Primary Key(s)
* Indicator to let downstream processes know to delete this row

## Change

With this PR, we will be including the before object if exists. [1] 

The improvements to be expected:

1. Rows that were deleted from the source but `topicConfigs[0].softDelete = true` will not have additional metadata
2. This will help with our subsequent errors around table partitioning

[1] - MongoDB by default does not emit the `before` object, to opt in: https://www.mongodb.com/docs/manual/reference/command/collMod/#change-streams-with-document-pre--and-post-images


Tang8330:
Converting this back into draft, going to refactor the tests based on #386

PR #387, opened 2024-03-25
[Typing] `parseField` to start emitting an error

Tang8330:
## Context

Previously, `parseField` would not emit an error and if the translation failed, it would just return the original value. However, this masked the actual translation error from the caller which makes it hard for other codepaths to reliably call this function.


nathan-artie (approved):
Love it!

PR #388, opened 2024-03-26
Fix Microsoft SQL Server booleans

Tang8330:
Solving for data indigestion.

Because Microsoft SQL Server doesn't support BOOLEANs, some data replication tools will create `BIGINT` as opposed to `BIT` columns.

Transfer then tries to load `boolean` values into a `BIGINT` column which fails the typecheck. We can safely load `boolean` values into a `BIT` column because the underlying Microsoft SQL driver expects and will cast `boolean` values into `BIT` values.

### DIFF
--- a/clients/mssql/values.go
+++ b/clients/mssql/values.go
@@ -88,12 +88,21 @@ func parseValue(colVal any, colKind columns.Column, additionalDateFmts []string)
 
 		return colVal, nil
 	case typing.Boolean.Kind:
-		// If it's already a boolean, return it. Else, convert it.
+		// We need to convert boolean into a bit for Microsoft SQL Server.
 		if val, isOk := colVal.(bool); isOk {
-			return val, nil
+			if val {
+				return 1, nil
+			} else {
+				return 0, nil
+			}
 		}
 
-		return strconv.ParseBool(colValString)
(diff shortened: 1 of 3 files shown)


clients/mssql/values.go
```diff
@@ -88,12 +88,21 @@ func parseValue(colVal any, colKind columns.Column, additionalDateFmts []string)
...
+			}
+		}
+
+		val, err := strconv.ParseBool(colValString)
+		if err != nil {
+			return nil, err
 		}
 
-		return strconv.ParseBool(colValString)
+		return parseValue(val, colKind, additionalDateFmts)
```
nathan-artie, line 105: It looks like `parseValue` will just return a boolean value if it's already a boolean. Do we just want to do:
```
if val {
				return 1, nil
			} else {
				return 0, nil
			}
```?

Tang8330: We didn't need to convert this as the Microsoft SQL Server driver will automatically parse booleans for BIT columns

Tang8330: In fact, by changing this to an integer, it broke the driver which expected the values to be booleans.

nathan-artie (approved):

PR #389, opened 2024-03-26
Remove nil check

nathan-artie:
Fixes `tautological condition: non-nil != nil`

### DIFF
--- a/lib/sql/escape.go
+++ b/lib/sql/escape.go
@@ -55,7 +55,7 @@ func EscapeName(name string, uppercaseEscNames bool, args *NameArgs) string {
 			name = strings.ToUpper(name)
 		}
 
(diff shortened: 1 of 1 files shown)


lib/sql/escape.go
```diff
@@ -55,7 +55,7 @@ func EscapeName(name string, uppercaseEscNames bool, args *NameArgs) string {
 			name = strings.ToUpper(name)
 		}
 
-		if args != nil && args.DestKind == constants.BigQuery {
```
nathan-artie, line 58: We have a nill check https://github.com/artie-labs/transfer/pull/389/files#diff-9bd9379792ed65d6ea87105277a8e4d5f488095afe33661b5ff27242e0cbe1bcR21

Tang8330 (approved):

PR #394, opened 2024-03-29
[Microsoft SQL Server] Exceeded values guardrail

Tang8330:
## Problem

Imagine a column in Microsoft SQL Server is `VARCHAR(255)`, if the actual value Artie is replicating over is greater than 255 the pipeline would fail.

### DIFF
--- a/clients/mssql/values.go
+++ b/clients/mssql/values.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 
 	"github.com/artie-labs/transfer/lib/config/constants"
-	"github.com/artie-labs/transfer/lib/stringutil"
 	"github.com/artie-labs/transfer/lib/typing"
 	"github.com/artie-labs/transfer/lib/typing/columns"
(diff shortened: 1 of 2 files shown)


clients/mssql/values.go
```diff
@@ -54,9 +53,12 @@ func parseValue(colVal any, colKind columns.Column, additionalDateFmts []string)
 			}
 
 			colValString = string(colValBytes)
-		} else {
-			// Else, make sure we escape the quotes.
-			colValString = stringutil.Wrap(colVal, true)
```
Tang8330, line 59: This is not needed.

nathan-artie: Nice!

nathan-artie: Do we need to insert a JSON exceeded value marker if the colVal is an array or JSON object?

Tang8330: I don't think so. This branch of logic already established that this column is string-based (or else it'd be in ARRAY or STRUCT).

PR #396, opened 2024-03-30
[Typing] Convert bool to bit if the column kind is an integer

Tang8330:
## Context

This PR adds the ability to cast a `boolean` value into a `bit` if the column in the destination is an `integer`.

I also rewrote the tests to cover every case and make it more clear what is being tested.

### DIFF
--- a/lib/typing/values/string_test.go
+++ b/lib/typing/values/string_test.go
@@ -5,18 +5,14 @@ import (
 	"testing"
 	"time"
 
-	"github.com/artie-labs/transfer/lib/ptr"
-	"github.com/artie-labs/transfer/lib/typing/decimal"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/artie-labs/transfer/lib/config/constants"
-
-	"github.com/artie-labs/transfer/lib/typing/ext"
-
+	"github.com/artie-labs/transfer/lib/ptr"
 	"github.com/artie-labs/transfer/lib/typing"
-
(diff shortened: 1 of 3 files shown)


lib/typing/values/string_test.go
```diff
@@ -1,289 +1,135 @@
 package values
 
 import (
-	"fmt"
```
Tang8330, line 4: Rewrote all the tests

lib/typing/values/string_test.go
```diff
@@ -16,274 +15,118 @@ import (
...
-			colVal: birthDateTime,
-			colKind: columns.Column{
-				KindDetails: dateTimeKind,
-			},
-			expectedString: "2022-09-06T03:19:24.942Z",
-		},
+	{
+		// Integer
+		// Floats first.
+		val, err := ToString(float32(45452.999991), columns.Column{KindDetails: typing.Integer}, nil)
```
nathan-artie, line 88: Should we be erroring in this case rather than supporting float -> int since it'll result in data loss?

Tang8330: There is precision loss as we're rounding up. A hard failure here doesn't necessarily make sense either as most customers would just create a float compatible column in the destination.

Tang8330: Another reason for this to happen in a legitimate cause could be that they are inconsistent about passing in the optional schema and JSON marshal made these values into `floats` but they're actually ints.

nathan-artie (approved):

PR #397, opened 2024-04-01
Minor jitter improvement

Tang8330:
Cap the attempts so we don't cause integer overflow

### DIFF
--- a/lib/jitter/sleep.go
+++ b/lib/jitter/sleep.go
@@ -1,6 +1,7 @@
 package jitter
 
 import (
+	"math"
 	"math/rand"
 	"time"
 )
@@ -19,9 +20,11 @@ func Jitter(baseMs, maxMs, attempts int) time.Duration {
 		return time.Duration(0)
 	}
 
-	// Cap the attempts to be 30 so we don't have integer overflows.
-	if attemptsMaxMs := baseMs * powerOfTwo(max(attempts, 30)); attemptsMaxMs > 0 {
-		maxMs = min(maxMs, attemptsMaxMs)
+	// maxThreshold is used to make sure the multiplier does not cause any value overflow.
+	// The 1 million comes from `ms` and `ns` conversion.
+	maxThreshold := math.MaxInt / baseMs / 1_000_000
+	if multiplier := powerOfTwo(min(attempts, 20)); multiplier <= maxThreshold {
+		maxMs = min(maxMs, baseMs*multiplier)
 	}
 
 	return time.Duration(rand.Intn(maxMs)) * time.Millisecond
--- a/lib/jitter/sleep.go
+++ b/lib/jitter/sleep.go
@@ -7,6 +7,10 @@ import (
 
 const DefaultMaxMs = 3500
 
+func powerOfTwo(n int) int {
+	return 1 << n
+}
+
 func Jitter(baseMs, maxMs, attempts int) time.Duration {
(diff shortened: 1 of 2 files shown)


lib/jitter/sleep.go
```diff
@@ -15,11 +19,10 @@ func Jitter(baseMs, maxMs, attempts int) time.Duration {
 		return time.Duration(0)
 	}
 
-	// Check for overflows when computing base * 2 ** attempts.
-	if attemptsMaxMs := baseMs * (1 << attempts); attemptsMaxMs > 0 {
+	// Cap the attempts to be 30 so we don't have integer overflows.
```
nathan-artie, line 22: - This would need to be `min` instead of `max`, otherwise we're going to use 30 as attempts when attempts is 1.
- Capping `attempts` doesn't change the behavior of this code as it already handles the overflows just fine: overflows will cause `attemptsMaxMs` to be negative and we'll use `maxMs` instead in that case). 
- if `baseMs` is >= `8589934592` `2** 30 * 8589934591` will still overflow on a 64-bit machine (and `baseMs` >= 2 will overflow on a 32-bit machine)
- If we want to explicitly check for overflows rather than rely on them producing negative numbers then we should do something like:
```
if x := powerOfTwo(min(attempts, 30)); math.MaxInt/baseMs <= x {
    // No overflow.
    maxMs = min(maxMs, baseMs*x)
}
```

Tang8330: Oops, yeah it should be min

lib/jitter/sleep.go
```diff
@@ -15,11 +20,12 @@ func Jitter(baseMs, maxMs, attempts int) time.Duration {
 		return time.Duration(0)
 	}
 
-	// Check for overflows when computing base * 2 ** attempts.
-	if attemptsMaxMs := baseMs * (1 << attempts); attemptsMaxMs > 0 {
-		maxMs = min(maxMs, attemptsMaxMs)
+	// maxThreshold is used to make sure the multiplier does not cause any value overflow.
+	// The 1 million comes from `ms` and `ns` conversion.
+	maxThreshold := math.MaxInt / baseMs / 1_000_000
+	if multiplier := powerOfTwo(min(attempts, 20)); multiplier <= maxThreshold {
```
nathan-artie, line 26: Where does 20 come from?

Tang8330: just a generous budget for 2^20

lib/jitter/sleep.go
```diff
@@ -15,11 +20,12 @@ func Jitter(baseMs, maxMs, attempts int) time.Duration {
 		return time.Duration(0)
 	}
 
-	// Check for overflows when computing base * 2 ** attempts.
-	if attemptsMaxMs := baseMs * (1 << attempts); attemptsMaxMs > 0 {
-		maxMs = min(maxMs, attemptsMaxMs)
+	// maxThreshold is used to make sure the multiplier does not cause any value overflow.
+	// The 1 million comes from `ms` and `ns` conversion.
+	maxThreshold := math.MaxInt / baseMs / 1_000_000
```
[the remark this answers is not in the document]
Tang8330: I'm down

lib/jitter/sleep.go
nathan-artie: Gotcha. It would be nice to add a docstring to `Jitter()` explaining the limitations since we're using it in `reader` and other projects.

PR #404, opened 2024-04-05
Cleaning up config validation

Tang8330:
## Changes

Main changes here are to clean up the verbosity of config validate

### DIFF
--- a/lib/config/config.go
+++ b/lib/config/config.go
@@ -20,7 +20,7 @@ const (
 	defaultFlushTimeSeconds = 10
 	defaultFlushSizeKb      = 25 * 1024 // 25 mb
 	defaultBufferPoolSize   = 30000
-	bufferPoolSizeStart     = 5
+	bufferPoolSizeMin       = 5
 
 	FlushIntervalSecondsMin = 5
 	FlushIntervalSecondsMax = 6 * 60 * 60
@@ -256,8 +256,8 @@ func (c Config) Validate() error {
 			c.FlushIntervalSeconds, FlushIntervalSecondsMin, FlushIntervalSecondsMax)
 	}
 
-	if bufferPoolSizeStart > int(c.BufferRows) {
-		return fmt.Errorf("buffer pool is too small, min value: %d, actual: %d", bufferPoolSizeStart, int(c.BufferRows))
+	if bufferPoolSizeMin > int(c.BufferRows) {
+		return fmt.Errorf("buffer pool is too small, min value: %d, actual: %d", bufferPoolSizeMin, int(c.BufferRows))
 	}
 
 	if !constants.IsValidDestination(c.Output) {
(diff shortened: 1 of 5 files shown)


lib/config/config.go
```diff
@@ -20,8 +20,8 @@ const (
 	defaultFlushTimeSeconds = 10
 	defaultFlushSizeKb      = 25 * 1024 // 25 mb
 
-	flushIntervalSecondsStart = 5
-	flushIntervalSecondsEnd   = 6 * 60 * 60
+	FlushIntervalSecondsStart = 5
```
Tang8330, line 23: I made these public so I can import them in our API server.

nathan-artie: It'd be a little clearer to call these `FlushIntervalSecondsMin` and `FlushIntervalSecondsMax`

lib/config/config.go
```diff
@@ -249,24 +248,20 @@ func (c Config) ValidateRedshift() error {
...
 	}
 
 	if bufferPoolSizeStart > int(c.BufferRows) {
-		return fmt.Errorf("config is invalid, buffer pool is too small, min value: %v, actual: %v", bufferPoolSizeStart, int(c.BufferRows))
+		return fmt.Errorf("buffer pool is too small, min value: %d, actual: %d", bufferPoolSizeStart, int(c.BufferRows))
 	}
 
 	if !constants.IsValidDestination(c.Output) {
-		return fmt.Errorf("config is invalid, output: %s is invalid", c.Output)
+		return fmt.Errorf("invalid output: %s", c.Output)
```
nathan-artie, line 264: `output` is super vague, maybe "invalid output destination"?

lib/numbers/numbers.go
```diff
@@ -1,12 +1,6 @@
...
-type BetweenEqArgs struct {
-	Start  int
-	End    int
-	Number int
-}
-
 // BetweenEq - Looks something like this. start <= number <= end
-func BetweenEq(args BetweenEqArgs) bool {
-	return args.Number >= args.Start && args.Number <= args.End
+func BetweenEq(start, end, number int) bool {
```
nathan-artie, line 4: If you use `[T ~int | ~int32 | ~int64]` this will support `int`, `int32`, and `int64`. Could even throw `~float32 | ~float64` in there.
```
// BetweenEq - Looks something like this. start <= number <= end
func BetweenEq[T ~int | ~int32 | ~int64](start, end, number T) bool {
	return number >= start && number <= end
}
```

nathan-artie (approved):

lib/numbers/numbers.go
Tang8330: Sure!

lib/config/config.go
Tang8330: Sgtm

Tang8330: sgtm

PR #405, opened 2024-04-05
[DDL] Minor refactor of the ddl pkg

Tang8330:
## Changes

1. Adding a debug log to dropping temporary tables
2. Changing `AlterTable` to a receiver method

### DIFF
--- a/clients/bigquery/bigquery.go
+++ b/clients/bigquery/bigquery.go
@@ -49,7 +49,7 @@ func (s *Store) PrepareTemporaryTable(tableData *optimization.TableData, tableCo
 			Mode:              tableData.Mode(),
 		}
 
-		if err := tempAlterTableArgs.Alter(tableData.ReadOnlyInMemoryCols().GetColumns()...); err != nil {
+		if err := tempAlterTableArgs.AlterTable(tableData.ReadOnlyInMemoryCols().GetColumns()...); err != nil {
 			return fmt.Errorf("failed to create temp table: %w", err)
 		}
 	}
--- a/clients/bigquery/bigquery.go
+++ b/clients/bigquery/bigquery.go
@@ -49,7 +49,7 @@ func (s *Store) PrepareTemporaryTable(tableData *optimization.TableData, tableCo
 			Mode:              tableData.Mode(),
 		}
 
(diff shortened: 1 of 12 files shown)


clients/bigquery/bigquery.go
```diff
@@ -49,7 +49,7 @@ func (s *Store) PrepareTemporaryTable(tableData *optimization.TableData, tableCo
 			Mode:              tableData.Mode(),
 		}
 
-		if err := ddl.AlterTable(tempAlterTableArgs, tableData.ReadOnlyInMemoryCols().GetColumns()...); err != nil {
+		if err := tempAlterTableArgs.Alter(tableData.ReadOnlyInMemoryCols().GetColumns()...); err != nil {
```
nathan-artie, line 52: I feel like dropping `Table` calling the function just `Alter` makes it hard to tell what is being altered, and now that it's a receiver function on an argument struct if I were seeing it for the first time I would assume that this code was just altering the `tempAlterTableArgs` struct, not creating a temporary table.

Tang8330: Oh interesting. For me, I look at `ALTER` and automatically assume it's the SQL keyword for DDL. I can add `AlterTable` back to be more clear though

lib/destination/ddl/ddl.go
```diff
@@ -23,7 +23,9 @@ import (
 // Temporary tables look like this: database.schema.tableName__artie__RANDOM_STRING(5)_expiryUnixTs
 func DropTemporaryTable(dwh destination.DataWarehouse, fqTableName string, shouldReturnError bool) error {
 	if strings.Contains(strings.ToLower(fqTableName), constants.ArtiePrefix) {
-		_, err := dwh.Exec(fmt.Sprintf("DROP TABLE IF EXISTS %s", fqTableName))
+		sqlCommand := fmt.Sprintf("DROP TABLE IF EXISTS %s", fqTableName)
+		slog.Debug("Dropping temporary table", slog.String("sql", sqlCommand))
+		_, err := dwh.Exec(sqlCommand)
```
nathan-artie, line 28: Could combine with the following line.

nathan-artie (approved):

lib/destination/ddl/ddl.go
Tang8330: Nice

PR #410, opened 2024-04-06
Simplify `parseField`

nathan-artie:
Added all the types from `typesThatRequireTypeCasting` to the `default` case in `parseField` so that `typesThatRequireTypeCasting` is no longer needed.


Tang8330 (approved):
Beautiful

PR #416, opened 2024-04-08
[debezium] Don't error log for int64 time values

nathan-artie:


Tang8330 (approved):

lib/debezium/types.go
```diff
@@ -134,12 +134,21 @@ func (f Field) ParseValue(value any) (any, error) {
...
-			// Since this value is coming from Kafka, and will have been marshaled to a JSON string, it should always
-			// be a float64. Let's check this if this assumption holds and if so clean up the code below so that we
-			// aren't doing float -> string -> float.
+
+		switch value.(type) {
+		case float64:
+			// This value is coming from Kafka, and will have been marshaled to a JSON string, so when it is
+			// unmarshalled it will be a float64
+			// -> Pass.
+		case int64:
```
Tang8330, line 143: Aren't we marshaling and unmarshaling from Reader before calling Transfer? Why is this happening?

nathan-artie: We are, I want to see if we can avoid having to do so since marshaling and unmarshaling are performance heavy operations and will reduce throughput.

PR #425, opened 2024-04-10
Pull out jitter upper bounds computation

nathan-artie:
This way we can test the upper bound values before they pass through `rand.Intn`.

### DIFF
--- a/lib/jitter/sleep.go
+++ b/lib/jitter/sleep.go
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-const DefaultMaxMs = int64(3500)
+const DefaultMaxMs = 3500
 
 func computeJitterUpperBoundMs(baseMs, maxMs, attempts int) int {
 	// https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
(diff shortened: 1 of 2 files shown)


lib/jitter/sleep.go
```diff
@@ -7,19 +7,26 @@ import (
...
 
-	ms := rand.Intn(maxMs)
-	return time.Duration(ms) * time.Millisecond
+	return maxMs
+}
+
+func Jitter(baseMs, maxMs, attempts int) time.Duration {
+	upperBoundMs := computeJitterUpperBoundMs(baseMs, maxMs, attempts)
+	if upperBoundMs <= 0 {
+		return time.Duration(0)
```
Tang8330, line 29: can we just return max ms here?

Tang8330 (approved):

lib/jitter/sleep.go
nathan-artie: I'm like 95% sure here that if we've made it to this line `maxMs` will be <= 0, but explicitly returning `time.Duration(0)` lets us avoid returning negative numbers if `maxMs` is negative.

PR #428, opened 2024-04-11
[Redshift] Improve Sweep query

Tang8330:
The current sweep query picks up sequences from the primary key for the temporary table. This PR makes it so that we filter on `relkind`.

Also changing `Failed to parse unix string` to log an error so that we can see it in Sentry.

### DIFF
--- a/lib/destination/ddl/expiry.go
+++ b/lib/destination/ddl/expiry.go
@@ -19,7 +19,7 @@ func ShouldDeleteFromName(name string) bool {
 func shouldDeleteUnix(unixString string) bool {
 	unix, err := strconv.Atoi(unixString)
 	if err != nil {
-		slog.Warn("Failed to parse unix string", slog.Any("err", err), slog.String("unixString", unixString))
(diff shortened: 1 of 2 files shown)


nathan-artie (approved):

lib/destination/ddl/expiry.go
```diff
@@ -19,7 +19,7 @@ func ShouldDeleteFromName(name string) bool {
 func shouldDeleteUnix(unixString string) bool {
 	unix, err := strconv.Atoi(unixString)
 	if err != nil {
-		slog.Warn("Failed to parse unix string", slog.Any("err", err), slog.String("unixString", unixString))
+		slog.Error("Failed to parse unix string", slog.Any("err", err), slog.String("unixString", unixString))
```
nathan-artie, line 22: Could move to `ShouldDeleteFromName` so that we can include the full table name.

Tang8330: True. Doing that

PR #431, opened 2024-04-11
[Relational Data] Include the Before object

Tang8330:
## Changes

Today, for our deletes - we do not pass the `before` object through. Which has worked fine, for hard deletes - we only need the primary keys and an indicator on whether the row has been deleted anyways.

However, if we want to use merge predicates which can speed up [merges](https://select.dev/posts/snowflake-merges#:~:text=This%20is%20known%20as%20dynamic%20pruning%2C%20because%20Snowflake%20is%20determining%20which%20micro%2Dpartitions%20to%20prune%20(avoid%20scanning)%20while%20during%20the%20query%20execution%20after%20it%20has%20read%20the%20values%20in%20the%20source%20table.), we'll need to be able to join on non-primary key values such as `created_at`.

This PR opens up the ability to pass through the before state for relational sources. Once this works well, we'll also add this for MongoDB.

### DIFF
--- a/lib/cdc/util/relational_event.go
+++ b/lib/cdc/util/relational_event.go
@@ -130,6 +130,7 @@ func (s *SchemaEventPayload) parseAndMutateMapInPlace(retMap map[string]any, kin
 			if val, parseErr := field.ParseValue(fieldVal); parseErr == nil {
 				retMap[field.FieldName] = val
 			} else {
+				// TODO: Make this a hard failure, confirm this with Datadog logs.
 				slog.Warn("Failed to parse field, using original value", slog.Any("err", parseErr),
 					slog.String("field", field.FieldName), slog.Any("value", fieldVal))
 			}
--- a/lib/cdc/util/relational_event.go
+++ b/lib/cdc/util/relational_event.go
@@ -121,17 +121,17 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]any, tc *kafkalib.TopicCon
 func (s *SchemaEventPayload) parseValues(retMap map[string]any, kind cdc.FieldLabelKind) map[string]any {
 	if schemaObject := s.Schema.GetSchemaFromLabel(kind); schemaObject != nil {
 		for _, field := range schemaObject.Fields {
-			_, isOk := retMap[field.FieldName]
+			fieldVal, isOk := retMap[field.FieldName]
 			if !isOk {
 				continue
 			}
 
-			val, parseErr := field.ParseValue(retMap[field.FieldName])
+			val, parseErr := field.ParseValue(fieldVal)
 			if parseErr == nil {
 				retMap[field.FieldName] = val
 			} else {
 				slog.Warn("Failed to parse field, using original value", slog.Any("err", parseErr),
-					slog.String("field", field.FieldName), slog.Any("value", retMap[field.FieldName]))
+					slog.String("field", field.FieldName), slog.Any("value", fieldVal))
 			}
 		}
 	}
--- a/lib/cdc/util/relational_event.go
+++ b/lib/cdc/util/relational_event.go
@@ -83,7 +83,7 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]any, tc *kafkalib.TopicCon
 	var retMap map[string]any
 	if len(s.Payload.After) == 0 {
 		if len(s.Payload.Before) > 0 {
-			retMap = s.Payload.Before
+			retMap = s.parseValues(s.Payload.Before, cdc.Before)
 		} else {
 			retMap = make(map[string]any)
 		}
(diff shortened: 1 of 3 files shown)


lib/cdc/util/relational_event.go
```diff
@@ -111,13 +113,16 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]any, tc *kafkalib.TopicCon
...
-	// Iterate over the schema and identify if there are any fields that require extra care.
-	afterSchemaObject := s.Schema.GetSchemaFromLabel(cdc.After)
-	if afterSchemaObject != nil {
-		for _, field := range afterSchemaObject.Fields {
+	return retMap
+}
+
+// parseValues will take `retMap` and `kind` (which part of the schema should we be inspecting) and then parse the values accordingly.
+// This will unpack any Debezium-specific values and convert them back into their original types.
+func (s *SchemaEventPayload) parseValues(retMap map[string]any, kind cdc.FieldLabelKind) map[string]any {
```
nathan-artie, line 121: Do we need to return anything since this is going to just mutate `retMap`?

nathan-artie: I feel like having this method return a value makes it seem like the `retMap` that is passed in isn't going to be changed.

Tang8330: I returned it because having the `retMap` be mutated in place seems brittle

Tang8330: And I didn't think we should create a new one since it'll increase the space complexity

lib/cdc/util/relational_event.go
```diff
@@ -111,13 +113,16 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]any, tc *kafkalib.TopicCon
...
+	if schemaObject := s.Schema.GetSchemaFromLabel(kind); schemaObject != nil {
+		for _, field := range schemaObject.Fields {
```
nathan-artie, line 123: On the next line we have the value of the field, but we're assigning it to `_`, we could assign it to a variable and then we wouldn't have to do the `retMap[field.FieldName]` lookup again.

lib/cdc/util/relational_event.go
nathan-artie: Isn't it going to be mutated in place though since `map[string]any` is going to be passed by reference and not value?

nathan-artie: Def agree that having it be mutated in place seems brittle.

Tang8330: It is, but we're not relying on it being updated in-place since it's being returned as well.

lib/cdc/util/relational_event.go
Tang8330: True.

lib/cdc/util/relational_event.go
nathan-artie: That seems a little dangerous to me though because it has an unexpected side effect that is not immediately obvious: that `s.Payload.Before`/ `s.Payload.After` will be modified after calling this method. I'd rather have it be obvious that the mutation is going to happen in place.

lib/cdc/util/relational_event.go
```diff
@@ -111,22 +113,25 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]any, tc *kafkalib.TopicCon
...
-				// Skipping b/c envelope mismatch with the actual request body
 				continue
 			}
 
-			val, parseErr := field.ParseValue(retMap[field.FieldName])
-			if parseErr == nil {
+			if val, parseErr := field.ParseValue(fieldVal); parseErr == nil {
 				retMap[field.FieldName] = val
 			} else {
 				slog.Warn("Failed to parse field, using original value", slog.Any("err", parseErr),
```
nathan-artie, line 133: I wonder if we can make this a hard error now.

lib/cdc/util/relational_event_test.go
```diff
@@ -130,37 +132,44 @@ func TestGetDataTestInsert(t *testing.T) {
...
-	_, isOk = evtData[tc.IdempotentKey]
-	assert.False(t, isOk, evtData)
+	kvMap := map[string]any{"pk": 1004}
+	{
+		// Postgres
+		var schemaEventPayload SchemaEventPayload
+		assert.NoError(t, json.Unmarshal([]byte(PostgresDelete), &schemaEventPayload))
+		assert.True(t, schemaEventPayload.DeletePayload())
+		data := schemaEventPayload.GetData(kvMap, tc)
+		for expectedKey, expectedValue := range expectedKeyValues {
```
nathan-artie, line 155: Will `assert.Equal(t, expectedKeyValues, data)` not work here?

nathan-artie (approved):

PR #432, opened 2024-04-12
Improving TableConfig code

Tang8330:
## Changes

* Rewriting certain tests
* Made `GetTableConfig` into a receiver method

### DIFF
--- a/clients/shared/table_config.go
+++ b/clients/shared/table_config.go
@@ -30,7 +30,7 @@ type GetTableCfgArgs struct {
 	DropDeletedColumns bool
 }
 
-func (g *GetTableCfgArgs) ShouldParseComment(comment string) bool {
+func (g GetTableCfgArgs) ShouldParseComment(comment string) bool {
 	if g.EmptyCommentValue != nil && comment == *g.EmptyCommentValue {
 		return false
 	}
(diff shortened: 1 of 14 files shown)


clients/shared/table_config.go
```diff
@@ -30,7 +30,7 @@ type GetTableCfgArgs struct {
 	DropDeletedColumns bool
 }
 
-func (g *GetTableCfgArgs) ShouldParseComment(comment string) bool {
```
Tang8330, line 33: ptr is unnecessary because the object isn't getting mutated.

lib/typing/snowflake.go
```diff
@@ -6,11 +6,9 @@ import (
...
 )
 
-// https://docs.snowflake.com/en/sql-reference/intro-summary-data-types.html
-
-func SnowflakeTypeToKind(snowflakeType string) KindDetails {
-	snowflakeType = strings.ToLower(snowflakeType)
-
+// snowflakeTypeToKind - Converts a Snowflake type to a KindDetails.
+// Following this spec: https://docs.snowflake.com/en/sql-reference/intro-summary-data-types.html
+func snowflakeTypeToKind(snowflakeType string) KindDetails {
```
nathan-artie, line 11: We can probably also get rid of the `strings.ToLower` here: https://github.com/artie-labs/transfer/pull/432/files#diff-e6f658cbda168f6e7a233f8bcf5807e9b37544c785787f3a1c4e3890c506b681R25

Tang8330: Good catch!

lib/typing/bigquery.go
```diff
@@ -9,9 +9,8 @@ import (
 
 const StreamingTimeFormat = "15:04:05"
 
-func BigQueryTypeToKind(rawBqType string) KindDetails {
+func bigQueryTypeToKind(rawBqType string) KindDetails {
 	bqType := rawBqType
-	bqType = strings.ToLower(bqType)
```
nathan-artie, line 14: Can probably remove the `strings.ToLower` call on line 33.

nathan-artie (approved):

PR #438, opened 2024-04-15
Adding MERGE Predicates

Tang8330:
Allowing merge predicate columns to be specified. We will then use it as part of the merge clause. 

For larger tables, this is beneficial because this helps reduce the amount of data required to be scanned when Transfer merges.


lib/destination/dml/merge.go
```diff
@@ -227,10 +227,10 @@ func (m *MergeArgument) GetStatement() (string, error) {
 	subQuery := fmt.Sprintf("( %s )", m.SubQuery)
 	if m.DestKind == constants.BigQuery {
 		subQuery = m.SubQuery
+	}
 
-		if len(m.AdditionalEqualityStrings) > 0 {
```
Tang8330, line 231: We only had this capability for BigQuery before, so pulling this out now.

clients/snowflake/writes.go
```diff
@@ -49,7 +50,16 @@ func (s *Store) Merge(tableData *optimization.TableData) error {
 			}
 		}
 
-		err = shared.Merge(s, tableData, s.config, types.MergeOpts{})
+		var additionalEqualityStrings []string
+		if len(tableData.TopicConfig.AdditionalMergeColumns) > 0 {
+			for _, additionalMergeCol := range tableData.TopicConfig.AdditionalMergeColumns {
+				additionalEqualityStrings = append(additionalEqualityStrings, fmt.Sprintf("c.%s = cc.%s", additionalMergeCol, additionalMergeCol))
```
Tang8330, line 56: c and cc are table aliases for staging and target table respectively.

lib/kafkalib/topic.go
```diff
@@ -47,6 +47,7 @@ type TopicConfig struct {
 	IncludeArtieUpdatedAt     bool                        `yaml:"includeArtieUpdatedAt"`
 	IncludeDatabaseUpdatedAt  bool                        `yaml:"includeDatabaseUpdatedAt"`
 	BigQueryPartitionSettings *partition.BigQuerySettings `yaml:"bigQueryPartitionSettings,omitempty"`
+	AdditionalMergeColumns    []string                    `yaml:"additionalMergeColumns,omitempty"`
```
Tang8330, line 50: When we go to Prod, we should create a separate type for this so we can embed additional information

clients/snowflake/writes.go
```diff
@@ -49,7 +50,16 @@ func (s *Store) Merge(tableData *optimization.TableData) error {
 			}
 		}
 
-		err = shared.Merge(s, tableData, s.config, types.MergeOpts{})
+		var additionalEqualityStrings []string
+		if len(tableData.TopicConfig.AdditionalMergeColumns) > 0 {
+			for _, additionalMergeCol := range tableData.TopicConfig.AdditionalMergeColumns {
```
nathan-artie, line 55: Does `additionalMergeCol` need to be escaped?

nathan-artie (approved):

clients/snowflake/writes.go
Tang8330: Yeah it does / should. Will add

danafallon (approved):

PR #440, opened 2024-04-15
[redshift] Implement `Dedupe`

nathan-artie:


clients/redshift/redshift.go
```diff
@@ -91,7 +92,34 @@ WHERE
 }
 
 func (s *Store) Dedupe(fqTableName string) error {
-	return fmt.Errorf("dedupe is not yet implemented")
+	stagingTableName := fmt.Sprintf("%s_dedupe_staging_%.5d", constants.ArtiePrefix, rand.Intn(100_000))
+
+	query := fmt.Sprintf(`
+BEGIN TRANSACTION;
```
Tang8330, line 98: Use the built-in transactions for go/sql? https://github.com/artie-labs/transfer/blob/a857a7fd9521bb14933270483279185444f81aa5/clients/shared/merge.go#L143

That way we can commit and rollback if there's an error

clients/redshift/redshift.go
```diff
@@ -91,7 +92,34 @@ WHERE
 }
 
 func (s *Store) Dedupe(fqTableName string) error {
-	return fmt.Errorf("dedupe is not yet implemented")
+	stagingTableName := fmt.Sprintf("%s_dedupe_staging_%.5d", constants.ArtiePrefix, rand.Intn(100_000))
```
Tang8330, line 95: Use this instead? https://github.com/artie-labs/transfer/blob/a857a7fd9521bb14933270483279185444f81aa5/clients/redshift/writes.go#L14

That way if this fails, our GC will drop the table

nathan-artie: Is it possible for this to fail in a way it doesn't get cleaned up though since it's in a transaction?

Tang8330: Probably not, still good to keep temp table names consistent though

clients/redshift/redshift.go
nathan-artie: Oh yeah def good call.

Tang8330 (approved):

clients/redshift/redshift.go
```diff
@@ -91,7 +93,47 @@ WHERE
...
+		stagingTableName,
+		// ;
+	)
+
+	transaction, err := s.Begin()
+	if err != nil {
+		return fmt.Errorf("failed to start transaction: %w", err)
+	}
+	defer func() {
+		if err := transaction.Rollback(); err != nil {
```
Tang8330, line 123: @nathan-artie Just reading through this code again, doesn't this mean that rollback will always run?

Tang8330: We can do something similar to this: https://github.com/artie-labs/transfer/blob/a857a7fd9521bb14933270483279185444f81aa5/clients/mssql/staging.go#L39

nathan-artie: It will, but from the [guide I was following](https://go.dev/doc/database/execute-transactions#:~:text=Defer%20the%20transaction%E2%80%99s%20rollback.%20If%20the%20transaction%20succeeds%2C%20it%20will%20be%20committed%20before%20the%20function%20exits%2C%20making%20the%20deferred%20rollback%20call%20a%20no%2Dop.%20If%20the%20transaction%20fails%20it%20won%E2%80%99t%20be%20committed%2C%20meaning%20that%20the%20rollback%20will%20be%20called%20as%20the%20function%20exits.):
> Defer the transaction’s rollback. If the transaction succeeds, it will be committed before the function exits, making the deferred rollback call a no-op. If the transaction fails it won’t be committed, meaning that the rollback will be called as the function exits.

I can add an check though to make it explicit.

Tang8330:
Good to know, yeah let's add it so it's more clear

clients/redshift/redshift.go
nathan-artie: Will make the change in a follow up PR, want to clean up the temporary table name logic so that it can be reused.

PR #442, opened 2024-04-16
Pass `TableIdentifier` to `DataWarehouse.Dedupe`

nathan-artie:
That way we can use it to generate a temporary table name for Redshift dedupes.

### DIFF
--- a/clients/redshift/redshift.go
+++ b/clients/redshift/redshift.go
@@ -90,7 +90,7 @@ WHERE
 	return shared.Sweep(s, tcs, queryFunc)
 }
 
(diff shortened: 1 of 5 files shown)


clients/redshift/redshift.go
```diff
@@ -90,7 +90,7 @@ WHERE
 	return shared.Sweep(s, tcs, queryFunc)
 }
 
-func (s *Store) Dedupe(fqTableName string) error {
+func (s *Store) Dedupe(tableData *optimization.TableData) error {
```
Tang8330, line 93: `_` ?

Tang8330 (approved):

PR #447, opened 2024-04-17
Parse Primary Keys.

Tang8330:
## Changes

We previously were not parsing primary keys.

### DIFF
--- a/lib/debezium/keys.go
+++ b/lib/debezium/keys.go
@@ -37,7 +37,7 @@ func (p PrimaryKeyPayload) parseAndReturnPayload() (map[string]any, error) {
 		} else {
 			parsedValue, err := p.Schema.Fields[idx].ParseValue(value)
 			if err != nil {
-				return nil, fmt.Errorf("failed to parse primary key: %w", err)
+				return nil, fmt.Errorf("failed to parse primary key: %q: %w", key, err)
 			}
 
 			retMap[key] = parsedValue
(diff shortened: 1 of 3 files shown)


lib/debezium/keys.go
```diff
@@ -70,8 +100,7 @@ func parsePartitionKeyStruct(keyBytes []byte) (map[string]any, error) {
 	}
 
 	var pkStruct map[string]any
-	err := json.Unmarshal(keyBytes, &pkStruct)
-	if err != nil {
+	if err := json.Unmarshal(keyBytes, &pkStruct); err != nil {
```
nathan-artie, line 103: It would be nice to change the error message on line 108 to something more specific since it's the same message as line 99.

lib/debezium/keys.go
```diff
@@ -3,20 +3,50 @@ package debezium
...
+		idx := slices.IndexFunc(p.Schema.Fields, func(field Field) bool {
+			return field.FieldName == key
+		})
+
+		if idx < 0 {
+			retMap[key] = value
+		} else {
+			parsedValue, err := p.Schema.Fields[idx].ParseValue(value)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse primary key: %w", err)
```
nathan-artie, line 40: Include key here?
`fmt.Errorf("failed to parse primary key %q: %w", key, err)`

nathan-artie (approved):

PR #448, opened 2024-04-17
Rewrite `keys_test.go`

Tang8330:


lib/debezium/keys_test.go
```diff
@@ -10,54 +10,85 @@ import (
...
+		assert.ErrorContains(t, err, `malformed key value pair: ""`)
+
+		_, err = parsePartitionKeyString([]byte("Struct{,,}"))
+		assert.ErrorContains(t, err, `malformed key value pair: ""`)
 	}
+	{
+		// Valid rows
+		kv, err := parsePartitionKeyString([]byte("Struct{hi=world,foo=bar}"))
+		assert.NoError(t, err)
+		assert.Equal(t, kv["hi"], "world")
```
nathan-artie, line 37: The order of arguments is backwards here.

nathan-artie: Ah if I turn on "show whitespace" then this line isn't actually modified.

nathan-artie (approved):

lib/debezium/keys_test.go
Tang8330: yeah, i just copied it over. i can change the order though

PR #450, opened 2024-04-17
Add a `TableIdentifier` for each data warehouse

nathan-artie:
Instead of having one `TableIdentifier`, we'll make `TableIdentifier` an interface and have one for each data warehouse.

PR #452, opened 2024-04-18
[tests] Split out `TempTableName` tests

nathan-artie:
We get import cycles if we try to import the concrete `DataWarehouse`s structs in `shared`, so I originally made a mock `DataWarehouse` to test `TempTableName` against, but I think it's better to just test on the actual `DataWarehouse` implementations and once this is merge it'll simplify https://github.com/artie-labs/transfer/pull/450.


clients/shared/temp_table_test.go
```diff
@@ -1,80 +0,0 @@
...
-		opts := optimization.FqNameOpts{MsSQLSchemaOverride: "mschema"}
-		tempTableName := TempTableName(MockDWH{kind: constants.MSSQL, opts: opts}, tableData, "sUfFiX")
-		assert.Equal(t, "mschema.table___artie_sUfFiX", trimTTL(tempTableName))
-	}
-	{
-		// Redshift:
-		tempTableName := TempTableName(MockDWH{kind: constants.Redshift}, tableData, "sUfFiX")
-		assert.Equal(t, "schema.table___artie_sUfFiX", trimTTL(tempTableName))
-	}
-	{
```
nathan-artie, line 70: S3 can't actually be used with `TempTableName` since it doesn't implement the `DataWarehouse` interface.

Tang8330 (approved):

PR #456, opened 2024-04-18
Log Debezium Type with Parse Error.

Tang8330:
So we know which case within the switch statement to look into

### DIFF
--- a/lib/cdc/util/relational_event.go
+++ b/lib/cdc/util/relational_event.go
@@ -129,8 +129,12 @@ func (s *SchemaEventPayload) parseAndMutateMapInPlace(retMap map[string]any, kin
 				retMap[field.FieldName] = val
 			} else {
 				// TODO: Make this a hard failure, confirm this with Datadog logs.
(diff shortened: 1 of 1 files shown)


lib/cdc/util/relational_event.go
```diff
@@ -129,8 +129,12 @@ func (s *SchemaEventPayload) parseAndMutateMapInPlace(retMap map[string]any, kin
 				retMap[field.FieldName] = val
 			} else {
 				// TODO: Make this a hard failure, confirm this with Datadog logs.
-				slog.Warn("Failed to parse field, using original value", slog.Any("err", parseErr),
-					slog.String("field", field.FieldName), slog.Any("value", fieldVal))
+				slog.Warn("Failed to parse field, using original value",
```
nathan-artie, line 132: We should also make this an error instead of a warning.

Tang8330: We could, it will blow up our Sentry though

nathan-artie (approved):

PR #459, opened 2024-04-18
[mssql] Pass `TableIdentifier` to `describeTableQuery`

nathan-artie:


clients/mssql/store.go
```diff
@@ -48,10 +48,14 @@ func (s *Store) Append(tableData *optimization.TableData) error {
 	})
 }
 
-func (s *Store) IdentifierFor(topicConfig kafkalib.TopicConfig, table string) types.TableIdentifier {
+func (s *Store) identifierFor(topicConfig kafkalib.TopicConfig, table string) TableIdentifier {
```
Tang8330, line 51: Why do we need a new function?

nathan-artie: The `IdentifierFor` method needs to return a `types.TableIdentifier` so that `Store` continues to implement the `DataWarehouse` interface. Unfortunately Go doesn't support returning supertypes in place of a subtypes (this is called covariance I just learned), so even though `TableIdentifier` (the MS SQL specific `TableIdentifier`) implements the `types.TableIdentifier` interface we are forced to return `types.TableIdentifier` instead of `TableIdentifier`.

Throughout `Store` we would prefer to use the concrete `TableIdentifier` type, so that we can access MS SQL specific attributes, so I added a private method that returns just that type. The alternative is to just have one `IdentifierFor` method and cast the returned type to `TableIdentifier`, but then we would have an error that we needed to handle. I would be fine panicking in that case though since it should never happen.

Tang8330: Ahh got it. Do you want to add that to the comments? It feels easily missable. 

Other alternative we can do is 
```go
	mssqlTableID, isOk := tableID.(TableIdentifier)
	if !isOk {
		return nil, fmt.Errorf("expected TableIdentifier, got %T", tableID)
	}
```

But it'll change the f(x) signature.

Tang8330 (approved):

clients/mssql/store.go
nathan-artie: I'll leave a comment, that way we can reuse it places where we don't necessarily want to have to return an error from.

PR #463, opened 2024-04-19
[debezium] Clean up `Field.ParseValue`

nathan-artie:
We haven't seen `Field type is %q but shouldCastToInt64 is false` messages in the logs which means that `Int16`/`Int32`/`Int64` `Field.Type`s are only used with an empty `Field.DebeziumType` or a `Field.DebeziumType` that is a time type, and in both cases we want to parse the value to `int64`.


lib/debezium/types.go
```diff
@@ -95,57 +95,19 @@ func toInt64(value any) (int64, error) {
 	return 0, fmt.Errorf("failed to cast value '%v' with type '%T' to int64", value, value)
 }
 
-func (f Field) shouldCastToInt64() bool {
-	if f.IsInteger() {
```
nathan-artie, line 99: `IsInteger` only returns true when `Field.DebeziumType` does not match one of our handled types AND `Field.Type` equaled `Int16`, `Int32`, or `Int64`.

lib/debezium/types.go
```diff
@@ -95,57 +95,19 @@ func toInt64(value any) (int64, error) {
...
-	case
-		Timestamp,
-		MicroTimestamp,
-		Date,
-		Time,
-		MicroTime,
-		DateKafkaConnect,
-		TimeKafkaConnect,
-		DateTimeKafkaConnect:
-		// We have verfied that these [Field.DebeziumType]s are always used with a [Field.Type] of [Int16], [Int32], or [Int64].
```
nathan-artie, line 113: `shouldCastToInt64` only returns true IF `Field.Type` equals `Int16`, `Int32`, or `Int64`.

lib/debezium/types.go
```diff
@@ -95,57 +95,19 @@ func toInt64(value any) (int64, error) {
...
-		// TODO: If we don't see the following error in the logs replace [Field.shouldCastToInt64] call with a simple
-		// check that [Field.Type] is [Int16], [Int32], or [Int64].
-		switch f.Type {
-		case Int16, Int32, Int64:
-			slog.Error(fmt.Sprintf("Field type is %q but shouldCastToInt64 is false", f.Type), slog.Any("value", value))
-		}
-	}
-
-	// Check if the field is an integer and requires us to cast it as such.
-	if f.IsInteger() {
```
nathan-artie, line 142: Move this to https://github.com/artie-labs/transfer/pull/463/files#diff-6f8a144d4abeb92f8240e55383e15b8323e83bb3b9f69088274ed61f4f344e96R155

We want to handle `DebeziumType`s first before we handle `Type`.

Tang8330 (approved):

PR #473, opened 2024-04-19
[BigQuery] Guardrail for empty JSON string.

Tang8330:
As per title.

### DIFF
--- a/clients/bigquery/cast.go
+++ b/clients/bigquery/cast.go
@@ -59,6 +59,12 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
 					colVal = fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder)
 				}
 			}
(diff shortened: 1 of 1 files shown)


clients/bigquery/cast.go
```diff
@@ -59,6 +59,12 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
 					colVal = fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder)
 				}
 			}
+
```
nathan-artie, line 62: Is the `colKind.KindDetails == typing.Struct` check above here redundant since we're switching on `colKind.KindDetails.Kind`?

nathan-artie (approved):

clients/bigquery/cast.go
Tang8330: Yeah seems like it. I'll rm

PR #484, opened 2024-04-22
[sql] Kill `NameArgs.Escape`

nathan-artie:
`Escape` is always true for non-test code and only ever false in test code.

### DIFF
--- a/lib/sql/escape.go
+++ b/lib/sql/escape.go
@@ -10,7 +10,6 @@ import (
 )
 
 type NameArgs struct {
-	Escape   bool
 	DestKind constants.DestinationKind
 }
 
@@ -18,7 +17,7 @@ type NameArgs struct {
 var symbolsToEscape = []string{":"}
 
 func EscapeNameIfNecessary(name string, uppercaseEscNames bool, args *NameArgs) string {
-	if args == nil || !args.Escape {
+	if args == nil {
 		return name
 	}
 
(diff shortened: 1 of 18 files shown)


lib/sql/escape.go
```diff
@@ -10,15 +10,14 @@ import (
...
 type NameArgs struct {
-	Escape   bool
 	DestKind constants.DestinationKind
 }
 
 // symbolsToEscape are additional keywords that we need to escape
 var symbolsToEscape = []string{":"}
 
 func EscapeNameIfNecessary(name string, uppercaseEscNames bool, args *NameArgs) string {
-	if args == nil || !args.Escape {
```
Tang8330, line 21: Is this safe? What about callers that expect the column to not be escaped?

nathan-artie: `args.Escape` is always true.

Tang8330: Cool

Tang8330: Do you want to kill NameArgs and just pass DestKind?

Tang8330 (approved):

lib/sql/escape.go
nathan-artie: I thought about that, and that's what I originally wanted to do, but it's a little tricky because sometimes `args` is null and it's annoying to create a pointer to `DestinationKind`. Maybe I'll do that anyways since it's a bit safer.

nathan-artie: Tried again and it ends up being a bigger lift than I want to do right now.

PR #501, opened 2024-04-24
[mssql] Escape all columns that do not start with "__artie"

nathan-artie:


lib/sql/escape.go
```diff
@@ -25,7 +25,7 @@ func NeedsEscaping(name string, destKind constants.DestinationKind) bool {
 	if destKind == constants.Redshift {
 		reservedKeywords = constants.RedshiftReservedKeywords
 	} else if destKind == constants.MSSQL {
-		reservedKeywords = constants.MSSQLReservedKeywords
+		return !strings.HasPrefix(name, constants.ArtiePrefix)
```
Tang8330, line 28: Why not just escape all?

Tang8330 (approved):

lib/sql/escape.go
nathan-artie: If we escape all we get test that say `artie delete flag doesn't exist`.

nathan-artie: I'm guessing there must be some places where we escape the column names and then check if the column names contain a specific value.

Tang8330: That's a guardrail in `dml/merge`

Tang8330: The comparison is running `"__artie_deleted` == __artie_deleted` which will always equate to false

PR #505, opened 2024-04-26
[Snowflake] Use Abort Detached Queries.

Tang8330:
## Context

Right now, for history mode / appends, it is susceptible to duplicates when there's a process restart.

The flow goes like this right now:

1. Create local stage and upload data to Snowflake
2. Load data into the target table via COPY INTO
3. Commit Kafka offset

However, if there's a process restart while we're in the midst of step 2, the connection will disconnect, but they query will still keep executing.

When the process then restarts, it will then fetch from the previous offset and keep going. This will result in duplicates.

## Solution

We're enabling [ABORT_DETACHED_QUERY](https://docs.snowflake.com/en/sql-reference/parameters#abort-detached-query) which should help reduce the amount of duplicates that arise from process restarts.

image

### DIFF
--- a/clients/snowflake/snowflake.go
+++ b/clients/snowflake/snowflake.go
@@ -97,6 +97,11 @@ func (s *Store) reestablishConnection() error {
 		return nil
 	}
 
+	params := make(map[string]*string)
+
+	// https://docs.snowflake.com/en/sql-reference/parameters#abort-detached-query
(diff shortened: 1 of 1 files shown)


Tang8330:
@danafallon @nathan-artie I was able to repro this with it on (default) and off. This does help reduce the amount of duplicate data. 

I'm going to publish a special image for this.

danafallon (approved):
🎉

PR #506, opened 2024-04-26
Adding yaml annotations to partitionFields

Tang8330:
Forgot to add YAML annotations from #505.

### DIFF
--- a/lib/kafkalib/partition/settings.go
+++ b/lib/kafkalib/partition/settings.go
@@ -22,9 +22,9 @@ type MergePredicates struct {
 }
 
 type BigQuerySettings struct {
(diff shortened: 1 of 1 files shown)


lib/kafkalib/partition/settings.go
```diff
@@ -18,7 +18,7 @@ var ValidPartitionBy = []string{
 }
 
 type MergePredicates struct {
-	PartitionField string
+	PartitionField string `yaml:"partitionField"`
```
danafallon, line 21: Does this need a json annotation too?

Tang8330: No it doesn't. I'm going to remove it from BigQuerySettings

danafallon (approved):

PR #511, opened 2024-04-29
[snowflake] Use `uppercaseEscNames: true` for tests

nathan-artie:
Will make the diff smaller when the time comes for https://github.com/artie-labs/transfer/pull/510/.


lib/destination/dml/merge_parts_test.go
```diff
@@ -31,7 +31,7 @@ type result struct {
 // getBasicColumnsForTest - will return you all the columns within `result` that are needed for tests.
 // * In here, we'll return if compositeKey=false - id (pk), email, first_name, last_name, created_at, toast_text (TOAST-able)
 // * Else if compositeKey=true - id(pk), email (pk), first_name, last_name, created_at, toast_text (TOAST-able)
-func getBasicColumnsForTest(compositeKey bool, uppercaseEscNames bool) result {
```
nathan-artie, line 34: Always false.

PR #514, opened 2024-04-29
[bigquery] Escape all columns that do not start with "__artie"

nathan-artie:
Since [BigQuery does not change the case of identifiers when they are not quoted](https://github.com/artie-labs/transfer/pull/489) we should be able to always quote column identifiers. This is based on https://github.com/artie-labs/transfer/pull/501 but BigQuery instead of MS SQL.

### DIFF
--- a/lib/sql/escape.go
+++ b/lib/sql/escape.go
@@ -22,7 +22,6 @@ func EscapeNameIfNecessary(name string, uppercaseEscNames bool, destKind constan
 
 func NeedsEscaping(name string, destKind constants.DestinationKind) bool {
 	var reservedKeywords []string
-
 	if destKind == constants.Redshift {
 		reservedKeywords = constants.RedshiftReservedKeywords
 	} else if destKind == constants.MSSQL || destKind == constants.BigQuery {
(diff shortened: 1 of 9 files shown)


lib/sql/escape.go
```diff
@@ -24,8 +24,10 @@ func NeedsEscaping(name string, destKind constants.DestinationKind) bool {
 	var reservedKeywords []string
 	if destKind == constants.Redshift {
 		reservedKeywords = constants.RedshiftReservedKeywords
-	} else if destKind == constants.MSSQL {
-		return !strings.HasPrefix(name, constants.ArtiePrefix)
+	} else if destKind == constants.MSSQL || destKind == constants.BigQuery {
+		if !strings.HasPrefix(name, constants.ArtiePrefix) {
```
nathan-artie, line 28: In the case that the name does start with `constants.ArtiePrefix`, I changed this so that don't immediately return so that we still run the logic for checking `symbolsToEscape`.

Tang8330: Why do that? There should never be reserved keywords that have the Artie prefix

nathan-artie: So that `__artie_foo:bar` still gets escaped.

Tang8330 (approved):

PR #516, opened 2024-04-30
Refactoring BigQuery Dedupe

Tang8330:
## Changes

Improving the way we handle deduping in BigQuery.

This way, we'll create a staging table that has all the duplicate rows.
* It stores the `latest` row in the staging table 
* We then delete all the duplicate rows from the target table
* We then copy the data from staging -> target
* Drop staging table

### DIFF
--- a/clients/bigquery/bigquery.go
+++ b/clients/bigquery/bigquery.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"strings"
+	"time"
 
 	"cloud.google.com/go/bigquery"
 	_ "github.com/viant/bigquery"
@@ -19,6 +21,8 @@ import (
 	"github.com/artie-labs/transfer/lib/logger"
 	"github.com/artie-labs/transfer/lib/optimization"
 	"github.com/artie-labs/transfer/lib/ptr"
+	"github.com/artie-labs/transfer/lib/stringutil"
+	"github.com/artie-labs/transfer/lib/typing"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


clients/bigquery/bigquery.go
```diff
@@ -143,10 +148,83 @@ func (s *Store) putTable(ctx context.Context, tableID types.TableIdentifier, row
...
 
-func (s *Store) Dedupe(tableID types.TableIdentifier, _ []string, _ kafkalib.TopicConfig) error {
-	fqTableName := tableID.FullyQualifiedName()
-	_, err := s.Exec(fmt.Sprintf("CREATE OR REPLACE TABLE %s AS SELECT DISTINCT * FROM %s", fqTableName, fqTableName))
-	return err
+func (s *Store) generateDedupeQueries(tableID, stagingTableID types.TableIdentifier, primaryKeys []string, topicConfig kafkalib.TopicConfig) []string {
+	var primaryKeysEscaped []string
+	for _, pk := range primaryKeys {
+		pkCol := columns.NewColumn(pk, typing.Invalid)
+		primaryKeysEscaped = append(primaryKeysEscaped, pkCol.Name(s.ShouldUppercaseEscapedNames(), s.Label()))
```
nathan-artie, line 155: Could do `sql.EscapeName(pk, false, s.Label())` since we always want escape BigQuery names.

clients/bigquery/bigquery.go
```diff
@@ -143,10 +148,83 @@ func (s *Store) putTable(ctx context.Context, tableID types.TableIdentifier, row
...
+	}
+
+	orderColsToIterate := primaryKeysEscaped
+	if topicConfig.IncludeArtieUpdatedAt {
+		orderColsToIterate = append(orderColsToIterate, constants.UpdateColumnMarker)
```
nathan-artie, line 160: I guess we probably want to escape `constants.UpdateColumnMarker` too.

clients/bigquery/bigquery.go
```diff
@@ -143,10 +148,83 @@ func (s *Store) putTable(ctx context.Context, tableID types.TableIdentifier, row
...
+	}
+
+	defer func() {
+		if !txCommitted {
+			if err = tx.Rollback(); err != nil {
+				slog.Warn("Failed to rollback tx", slog.Any("err", err))
+			}
+		}
+
+		_ = ddl.DropTemporaryTable(s, stagingTableID.FullyQualifiedName(), false)
```
nathan-artie, line 213: Do we only need to do this if the transaction was committed?

Tang8330: The transaction doesn't drop the table

Tang8330: Note the SQL text doesn't explicitly call this a temporary table. Temp tables in BQ behaves differently from Snowflake and Redshift.

Tang8330: They're regular tables that can have a TTL. This query drops the table so we don't need to wait for GC

clients/bigquery/bigquery.go
Tang8330: We can escape it, but we don't need to.

Tang8330: I could have it call `sql.EscapeNameIfNecessary` so that it'll get escaped once we address escaping Artie cols?

clients/snowflake/snowflake.go
```diff
@@ -131,13 +130,12 @@ func (s *Store) reestablishConnection() error {
 func (s *Store) generateDedupeQueries(tableID, stagingTableID types.TableIdentifier, primaryKeys []string, topicConfig kafkalib.TopicConfig) []string {
 	var primaryKeysEscaped []string
 	for _, pk := range primaryKeys {
-		pkCol := columns.NewColumn(pk, typing.Invalid)
-		primaryKeysEscaped = append(primaryKeysEscaped, pkCol.Name(s.ShouldUppercaseEscapedNames(), s.Label()))
+		primaryKeysEscaped = append(primaryKeysEscaped, sql.EscapeNameIfNecessary(pk, s.ShouldUppercaseEscapedNames(), s.Label()))
```
Tang8330, line 133: @nathan-artie Heads up - I escaped this too

clients/bigquery/bigquery.go
nathan-artie: Ah that was my guess (that the transaction doesn't drop the table).

clients/bigquery/bigquery.go
nathan-artie: Yep that works!

clients/snowflake/snowflake.go
nathan-artie: Nice!

nathan-artie (approved):

PR #517, opened 2024-04-30
[redshift] Escape all columns that do not start with "__artie"

nathan-artie:


lib/config/constants/redshift.go
```diff
@@ -1,166 +1,3 @@
 package constants
 
-// RedshiftReservedKeywords https://docs.aws.amazon.com/redshift/latest/dg/r_pg_keywords.html
-var RedshiftReservedKeywords = []string{
```
Tang8330, line 4: Nice

Tang8330 (approved):

PR #519, opened 2024-04-30
[snowflake] Escape all columns that do not start with "__artie"

nathan-artie:
With the caveat that `UppercaseEscapedNames` is set to true.

### DIFF
--- a/lib/sql/escape.go
+++ b/lib/sql/escape.go
@@ -31,6 +31,9 @@ func NeedsEscaping(name string, uppercaseEscNames bool, destKind constants.Desti
 		return false
 	case constants.Snowflake:
 		if uppercaseEscNames {
+			// If uppercaseEscNames is true then we will escape all identifiers that do not start with the Artie priefix.
+			// Since they will be uppercased afer they are escaped then they will result in the same value as if we
+			// we were to use them in a query without any escaping at all.
 			// TODO: Escape names that start with [constants.ArtiePrefix].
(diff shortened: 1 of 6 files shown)


lib/sql/escape.go
```diff
@@ -58,7 +69,7 @@ func EscapeName(name string, uppercaseEscNames bool, destKind constants.Destinat
 				slog.Bool("uppercaseEscapedNames", uppercaseEscNames),
 			)
 		}
-  } else if destKind == constants.Redshift {
+	} else if destKind == constants.Redshift {
```
nathan-artie, line 72: I think this is just a whitespace change.

Tang8330 (approved):

lib/sql/escape.go
nathan-artie: Probably from a merge conflict that was fixed in the GitHub UI.

lib/sql/escape.go
```diff
@@ -14,23 +14,39 @@ import (
...
 		if !strings.HasPrefix(name, constants.ArtiePrefix) {
 			return true
 		}
-	default:
-		if slices.Contains(constants.ReservedKeywords, name) {
-			return true
+	case constants.S3:
+		return false
+	case constants.Snowflake:
+		if uppercaseEscNames {
```
Tang8330, line 33: Want to leave a comment block here to explain the intent?

PR #524, opened 2024-05-01
Add `Dialect()` to `DataWarehouse`

nathan-artie:


Tang8330 (commented):
Can we combine these so we don't need 2 impls?

https://github.com/artie-labs/transfer/blob/98135da1914fdd9aa2888bb8a7cb1ffc0c39a832/lib/sql/escape.go#L16

Tang8330 (approved):

nathan-artie:
> Can we combine these so we don't need 2 impls?
> 
> https://github.com/artie-labs/transfer/blob/98135da1914fdd9aa2888bb8a7cb1ffc0c39a832/lib/sql/escape.go#L16

`dialectFor` should go away completely once I've changed everything to reference `s.Dialect()`.

PR #529, opened 2024-05-01
Pass `Dialect` down

nathan-artie:


lib/destination/dml/merge.go
```diff
@@ -103,7 +103,7 @@ func (m *MergeArgument) GetParts() ([]string, error) {
 		equalitySQLParts = append(equalitySQLParts, equalitySQL)
 	}
 
-	cols := m.Columns.GetEscapedColumnsToUpdate(*m.UppercaseEscNames, m.DestKind)
+	cols := m.Columns.GetEscapedColumnsToUpdate(m.Dialect)
```
Tang8330, line 106: We can remove `UppercaseEscNames` now I think

nathan-artie: Yep, I'm hoping so, was gonna do that in a follow up PR.

Tang8330 (approved):

lib/destination/dml/merge.go
Tang8330: Cool sgtm

PR #536, opened 2024-05-02
Pass `Dialect` to `Column.DefaultValue`

nathan-artie:


clients/shared/utils.go
```diff
@@ -25,7 +25,7 @@ func BackfillColumn(cfg config.Config, dwh destination.DataWarehouse, column col
 	}
 
 	additionalDateFmts := cfg.SharedTransferConfig.TypingSettings.AdditionalDateFormats
-	defaultVal, err := column.DefaultValue(&columns.DefaultValueArgs{Escape: true, DestKind: dwh.Label()}, additionalDateFmts)
+	defaultVal, err := column.DefaultValue(dwh.Dialect(), additionalDateFmts)
```
nathan-artie, line 28: This is the only place we call `DefaultValue` so I removed the `Escape` argument.

lib/typing/columns/default.go
```diff
@@ -12,28 +12,23 @@ import (
...
-		switch args.DestKind {
-		case constants.BigQuery:
+		switch dialect.(type) {
+		case sql.BigQueryDialect:
 			return "JSON" + stringutil.Wrap(c.defaultValue, false), nil
-		case constants.Redshift:
+		case sql.RedshiftDialect:
 			return fmt.Sprintf("JSON_PARSE(%s)", stringutil.Wrap(c.defaultValue, false)), nil
-		case constants.Snowflake:
+		case sql.SnowflakeDialect:
```
nathan-artie, line 31: @Tang8330 it doesn't seem like we're doing anything for MS SQL here, do we need a case for it?

Tang8330: Yeah we're not. We don't support backfill default values in MSSQL yet.

Do you want to leave a comment?

PR #539, opened 2024-05-02
Refactor Redshift Append

Tang8330:
## Changes

Instead of creating a separate staging table and then running `ALTER TABLE APPEND`, we can just directly load this data into the Redshift table.

By doing this, we don't have to worry about the edge cases of ALTER TABLE APPEND like having distribution mode clashes.


Tang8330:
@nathan-artie This unblocks the append mode for reader. We just need to rewrite the dedupe query and we can enable it for Redshift deploys.

clients/redshift/writes.go
```diff
@@ -1,27 +1,14 @@
...
-	temporaryTableID := shared.TempTableID(tableID, tableData.TempTableSuffix())
-	if err := shared.Append(s, tableData, types.AppendOpts{TempTableID: temporaryTableID}); err != nil {
-		return err
-	}
-
-	_, err := s.Exec(
-		fmt.Sprintf(`ALTER TABLE %s APPEND FROM %s;`, tableID.FullyQualifiedName(), temporaryTableID.FullyQualifiedName()),
-	)
-	return err
+	return shared.Append(s, tableData, types.AppendOpts{TempTableID: tableID})
```
nathan-artie, line 11: Is this comment outdated now: https://github.com/artie-labs/transfer/blob/415319168ed070283228491683db5067620ebf56/lib/destination/types/types.go#L48

nathan-artie (approved):

nathan-artie (approved):

PR #541, opened 2024-05-02
[snowflake] Always escape column identifiers

nathan-artie:
Adds a new flag `LegacyMode` to `SnowflakeDialect`. In legacy mode reserved column identifiers are lowercased and non-reserved column identifiers are uppercased. In regular mode all column identifiers are uppercased.

### DIFF
--- a/clients/snowflake/tableid.go
+++ b/clients/snowflake/tableid.go
@@ -7,7 +7,7 @@ import (
 	"github.com/artie-labs/transfer/lib/sql"
 )
 
-var dialect = sql.SnowflakeDialect{UppercaseEscNames: true}
+var dialect = sql.SnowflakeDialect{}
 
 type TableIdentifier struct {
 	database string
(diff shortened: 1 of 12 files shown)


clients/snowflake/tableid.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/artie-labs/transfer/lib/sql"
 )
 
-var dialect = sql.SnowflakeDialect{UppercaseEscNames: true}
+var dialect = sql.SnowflakeDialect{}
```
Tang8330, line 10: This isn't using legacy mode because we don't need to escape any more tables right?

Tang8330 (approved):

clients/snowflake/tableid.go
nathan-artie: Yep, we've been escaping all Snowflake tables for some time.

PR #555, opened 2024-05-03
[dml] Reuse result of `ValidColumns` call

nathan-artie:


lib/destination/dml/columns.go
```diff
@@ -22,10 +22,6 @@ func quoteColumns(cols []columns.Column, dialect sql.Dialect) []string {
 func buildColumnsUpdateFragment(columns []columns.Column, dialect sql.Dialect, skipDeleteCol bool) string {
 	var cols []string
 	for _, column := range columns {
-		if column.ShouldSkip() {
```
nathan-artie, line 25: Since `ValidColumns` does not include invalid columns we don't need to check `ShouldSkip()` anymore.

Tang8330 (approved):

PR #572, opened 2024-05-07
Upgrading MongoDB Dependency

Tang8330:
We are currently on a version from July 2023. 

Reading through the release notes, there should not be any breaking changes. However, they did deprecate certain methods that we relied on, particularly in `bson.go`. 

I followed the examples from this PR, which introduces the deprecation warnings: https://github.com/mongodb/mongo-go-driver/pull/1251

### DIFF
--- a/lib/typing/mongo/bson.go
+++ b/lib/typing/mongo/bson.go
@@ -1,6 +1,7 @@
 package mongo
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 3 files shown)


Tang8330:
@nathan-artie Heads up - I refactored some `bson.go` files as per their migration guide.

lib/typing/mongo/bson.go
```diff
@@ -30,21 +31,36 @@ func JSONEToMap(val []byte) (map[string]any, error) {
...
 	var bsonDoc bson.D
 	err := bson.UnmarshalExtJSON(val, false, &bsonDoc)
 	if err != nil {
 		return nil, fmt.Errorf("failed to unmarshal ext json: %w", err)
 	}
 
-	bytes, err := bson.MarshalExtJSONWithRegistry(createCustomRegistry().Build(),
-		bsonDoc, false, true)
+	buf := new(bytes.Buffer)
+	vw, err := bsonrw.NewExtJSONValueWriter(buf, false, true)
```
nathan-artie, line 41: Could move the buffer, valuewriter, and encoder into `createCustomRegistry()` and rename it `createExtJSONEncoder()` or something. I also wonder if we'd get any performance gains by creating an encoder once and reusing it.

nathan-artie: I guess you'd have to pass the buffer into that method though.

nathan-artie (approved):

PR #576, opened 2024-05-08
Add `ExecStatements` function

nathan-artie:


lib/destination/dwh.go
```diff
@@ -35,3 +37,37 @@ type Baseline interface {
...
+		return fmt.Errorf("statements is empty")
+	case 1:
+		slog.Debug("Executing...", slog.String("query", statements[0]))
+		if _, err := dwh.Exec(statements[0]); err != nil {
+			return fmt.Errorf("failed to execute statement: %w", err)
+		}
+
+		return nil
+	default:
+		tx, err := dwh.Begin()
```
Tang8330, line 55: What about rolling back tx if it failed?

nathan-artie: I can add that, I was just duplicating what we were already doing for Redshift merges https://github.com/artie-labs/transfer/blob/e275e192b8a886f5d8c4498da964ce37b85dd062/clients/shared/merge.go#L147-L168

Tang8330: Yeah, we should add it. We're also inconsistent about rolling it back.

lib/destination/dwh.go
```diff
@@ -35,3 +37,46 @@ type Baseline interface {
...
+		var committed bool
```
Tang8330, line 56: nit: bring committed right before the defer

lib/destination/dwh.go
```diff
@@ -35,3 +37,46 @@ type Baseline interface {
...
+		if err != nil {
+			return fmt.Errorf("failed to start tx: %w", err)
+		}
+		defer func() {
+			if !committed {
+				if err = tx.Rollback(); err != nil {
```
Tang8330, line 62: Use a different variable? This might shadow the actual `err`

Tang8330 (approved):
lgtm % c

PR #581, opened 2024-05-08
Rewriting Redshift Dedupe

Tang8330:
## Changes

This PR improves our Redshift deduplication logic. 

As part of this, I've also added a separate function to `TableIdentifier` to escape just the table name, since we need this for Redshift.

Once this is merged, I'm also going to remove all of our mocks.


clients/bigquery/tableid.go
```diff
@@ -31,6 +31,10 @@ func (ti TableIdentifier) Dataset() string {
 	return ti.dataset
 }
 
+func (ti TableIdentifier) EscapedTable() string {
```
nathan-artie, line 34: We should probably call this function in `FullyQualifiedName` too.

clients/bigquery/tableid_test.go
```diff
@@ -23,3 +23,11 @@ func TestTableIdentifier_FullyQualifiedName(t *testing.T) {
 	// Table name that is a reserved word:
 	assert.Equal(t, "`project`.`dataset`.`table`", NewTableIdentifier("project", "dataset", "table").FullyQualifiedName())
 }
+
+func TestTableIdentifier_EscapedTable(t *testing.T) {
+	// Table name that is not a reserved word:
```
nathan-artie, line 28: I don't think we need to test for reserved words anymore since we quote everything.

clients/s3/tableid.go
```diff
@@ -24,6 +24,10 @@ func (ti TableIdentifier) Schema() string {
 	return ti.schema
 }
 
+func (ti TableIdentifier) EscapedTable() string {
+	return ti.table
```
nathan-artie, line 28: Add a comment that we don't escape S3 tables?

clients/redshift/redshift.go
```diff
@@ -114,53 +114,64 @@ WHERE
...
-		// ; INSERT INTO
-		fqTableName,
-		// SELECT * FROM
-		stagingTableName,
-		// ; DROP TABLE
-		stagingTableName,
-		// ;
-	)
+func (s *Store) generateDedupeQueries(tableID, stagingTableID types.TableIdentifier, primaryKeys []string, topicConfig kafkalib.TopicConfig) []string {
+	var primaryKeysEscaped []string
```
nathan-artie, line 118: Can do `primaryKeysEscaped := sql.QuoteIdentifiers(primaryKeys, s.Dialect())`

clients/bigquery/tableid.go
Tang8330: sg, done

nathan-artie (approved):

PR #582, opened 2024-05-09
Removing mocks

Tang8330:
As per title.

* Adding mocks to `.gitignore` properly
* Generating them before we run tests


.gitignore
```diff
@@ -17,3 +19,6 @@ scratch/
 # .tfstate files
 *.tfstate
 *.tfstate.*
+
```
nathan-artie, line 22: Extra space?

nathan-artie (approved):

.gitignore
Tang8330: oops

PR #598, opened 2024-05-09
[ddl] Pass `Dialect` in `AlterTableArgs`

nathan-artie:
Since the  Store/DataWarehouse is now only needed to execute the queries we'll pass it in the `AlterTable` method call.

### DIFF
--- a/lib/destination/ddl/ddl.go
+++ b/lib/destination/ddl/ddl.go
@@ -52,6 +52,10 @@ type AlterTableArgs struct {
 }
 
 func (a AlterTableArgs) Validate() error {
+	if a.Dialect == nil {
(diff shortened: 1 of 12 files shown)


lib/destination/ddl/ddl.go
```diff
@@ -37,8 +37,8 @@ func DropTemporaryTable(dwh destination.DataWarehouse, fqTableName string, shoul
 }
 
 type AlterTableArgs struct {
-	Dwh destination.DataWarehouse
-	Tc  *types.DwhTableConfig
+	Dialect sql.Dialect
```
Tang8330, line 40: Validate to check if Dialect is nil?

Tang8330 (approved):

PR #610, opened 2024-05-10
[sql] Add `SupportsColumnKeyword` method to `Dialect`

nathan-artie:


lib/destination/ddl/ddl.go
```diff
@@ -137,11 +137,11 @@ func (a AlterTableArgs) buildStatements(cols ...columns.Column) ([]string, []col
 	} else {
 		for _, colSQLPart := range colSQLParts {
 			var sqlQuery string
-			if _, ok := a.Dialect.(sql.MSSQLDialect); ok {
+			if a.Dialect.SupportsColumnKeyword() {
```
Tang8330: Can we create an interface called `GetAlterColumnQuery()`?

Tang8330 (approved):

lib/destination/ddl/ddl.go
nathan-artie: Yep.

PR #611, opened 2024-05-10
Reload the client upon a rebalance

Tang8330:


lib/kafkalib/consumer.go
```diff
@@ -10,3 +12,31 @@ type Consumer interface {
...
+func ShouldReload(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// [27] Rebalance In Progress: the coordinator has begun rebalancing the group, the client should rejoin the group
+	return errors.Is(err, kafka.RebalanceInProgress)
+}
+
+func (r *Reader) Reload() error {
```
nathan-artie, line 37: This never returns an error.

processes/consumer/kafka.go
```diff
@@ -101,12 +101,19 @@ func StartConsumer(ctx context.Context, cfg config.Config, inMemDB *models.Datab
...
 			for {
 				kafkaMsg, err := kafkaConsumer.FetchMessage(ctx)
 				if err != nil {
+					if kafkalib.ShouldReload(err) {
+						if err = kafkaConsumer.Reload(); err != nil {
+							logger.Fatal("Failed to reload kafka consumer", slog.Any("err", err))
+						}
+					}
+
 					slog.With(artie.KafkaMsgLogFields(kafkaMsg)...).Warn("Failed to read kafka message", slog.Any("err", err))
```
nathan-artie, line 115: Put this in an `else` since if we reload we don't want to log?

processes/consumer/kafka.go
```diff
@@ -101,12 +101,19 @@ func StartConsumer(ctx context.Context, cfg config.Config, inMemDB *models.Datab
...
 			}
 
-			kafkaConsumer := kafka.NewReader(kafkaCfg)
+			kafkaConsumer := kafkalib.NewReader(kafkaCfg)
 			topicToConsumer.Add(topic, kafkaConsumer)
 			for {
 				kafkaMsg, err := kafkaConsumer.FetchMessage(ctx)
 				if err != nil {
+					if kafkalib.ShouldReload(err) {
+						if err = kafkaConsumer.Reload(); err != nil {
```
nathan-artie: Might want to log here that we're going to reload.

lib/kafkalib/consumer.go
Tang8330: thanks

processes/consumer/kafka.go
Tang8330: yeah just added

nathan-artie (approved):

lib/kafkalib/consumer.go
```diff
@@ -10,3 +12,31 @@ type Consumer interface {
...
+
+func NewReader(config kafka.ReaderConfig) *Reader {
+	return &Reader{
+		Reader: kafka.NewReader(config),
+		config: config,
+	}
+}
+
+func ShouldReload(err error) bool {
+	if err == nil {
```
nathan-artie, line 29: This could go away, `errors.Is` will return `false` for `nil`.

nathan-artie: https://go.dev/play/p/547vdvNWgJQ

Tang8330: ill do that tomorrow and also add a test

PR #615, opened 2024-05-10
Adding jitter sleep when establishing new Kafka connections

Tang8330:
As per title

### DIFF
--- a/processes/consumer/kafka.go
+++ b/processes/consumer/kafka.go
@@ -92,9 +92,7 @@ func StartConsumer(ctx context.Context, cfg config.Config, inMemDB *models.Datab
 	var wg sync.WaitGroup
 	for num, topic := range topics {
 		// It is recommended to not try to establish a connection all at the same time, which may overwhelm the Kafka cluster.
-		sleepDur := jitter.Jitter(100, 3000, num)
-		time.Sleep(sleepDur * time.Millisecond)
-
+		time.Sleep(jitter.Jitter(100, 3000, num) * time.Millisecond)
 		wg.Add(1)
 		go func(topic string) {
 			defer wg.Done()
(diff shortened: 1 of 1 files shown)


nathan-artie (approved):

processes/consumer/kafka.go
```diff
@@ -89,7 +90,9 @@ func StartConsumer(ctx context.Context, cfg config.Config, inMemDB *models.Datab
 	}
 
 	var wg sync.WaitGroup
-	for _, topic := range topics {
+	for num, topic := range topics {
+		// It is recommended to not try to establish a connection all at the same time, which may overwhelm the Kafka cluster.
+		time.Sleep(jitter.Jitter(100, 3000, num))
```
nathan-artie, line 95: Actually I wonder if we want to be passing in `num` here.

nathan-artie, line 95: I don't think we want to use `Jitter` here because it's an exponential algorithm and we probably want something that's more linear. For instance, when num is `4` the upper bound will be 1600ms but `5` and higher will produce `3000`, so then everything will essentially wait 3 seconds and then try to connect at the same time.

nathan-artie: Maybe `rand.Int63(100 * n)`.

Tang8330: What do you suggest?

Tang8330: `(200 * num) * time.Millisecond`?

nathan-artie: Although I suppose that is just the upper bound so it'll be somewhat random within that.

nathan-artie: >so then everything will essentially wait 3 seconds and then try to connect at the same time.
is not accurate.

Tang8330: Yeah it's not true exponential, it has randomness

```
random_between(0, min(cap, base * 2 ** attempt))
```

PR #630, opened 2024-05-13
Remove deleted column marker in append mode

nathan-artie:


clients/shared/append.go
```diff
@@ -21,10 +21,12 @@ func Append(dwh destination.DataWarehouse, tableData *optimization.TableData, op
...
 
+	cols, _ := columns.RemoveDeleteColumnMarker(tableConfig.Columns().GetColumns())
+
 	// We don't care about srcKeysMissing because we don't drop columns when we append.
 	_, targetKeysMissing := columns.Diff(
-		tableData.ReadOnlyInMemoryCols(),
-		tableConfig.Columns(),
+		tableData.ReadOnlyInMemoryCols().GetColumns(),
+		cols,
 		tableData.TopicConfig().SoftDelete,
```
Tang8330, line 30: Can we hardcode `softDelete` here to be false? It should never be a soft delete for append

Tang8330 (approved):

clients/shared/append.go
```diff
@@ -21,10 +21,12 @@ func Append(dwh destination.DataWarehouse, tableData *optimization.TableData, op
 		return fmt.Errorf("failed to get table config: %w", err)
 	}
 
+	cols, _ := columns.RemoveDeleteColumnMarker(tableConfig.Columns().GetColumns())
+
 	// We don't care about srcKeysMissing because we don't drop columns when we append.
 	_, targetKeysMissing := columns.Diff(
-		tableData.ReadOnlyInMemoryCols(),
-		tableConfig.Columns(),
+		tableData.ReadOnlyInMemoryCols().GetColumns(),
```
Tang8330, line 28: Actually, shouldn't we exclude it from the source and not target?

Isn't this going to add a column `__artie_deleted` in the destination?

clients/shared/append.go
nathan-artie: Yep!

clients/shared/append.go
nathan-artie: Ooh yeah good catch!

PR #644, opened 2024-05-14
Move `DefaultValue` to `shared`

nathan-artie:
Also combine the tests that were split in https://github.com/artie-labs/transfer/pull/636.


lib/sql/columns.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/artie-labs/transfer/lib/typing"
+	"github.com/artie-labs/transfer/lib/typing/columns"
+)
+
+func QuoteColumns(cols []columns.Column, dialect Dialect) []string {
```
nathan-artie, line 11: No changes just moved these two functions.

PR #649, opened 2024-05-14
[dml] Kill `MergeArgument`

nathan-artie:


Tang8330 (approved):

lib/sql/dialect.go
```diff
@@ -31,9 +31,12 @@ type Dialect interface {
 		subQuery string,
 		idempotentKey string,
 		primaryKeys []columns.Column,
+		// additionalEqualityStrings is used for handling BigQuery partitioned table merges
```
Tang8330, line 34: It's used for Snowflake as well

PR #655, opened 2024-05-14
[sql] Add constants for table aliases

nathan-artie:


clients/bigquery/dialect/dialect.go
```diff
@@ -15,9 +15,14 @@ import (
 	"github.com/artie-labs/transfer/lib/typing/ext"
 )
 
-const BQStreamingTimeFormat = "15:04:05"
+const (
+	srcAlias  = "cc"
+	destAlias = "c"
```
Tang8330, line 20: Can we rename this:

srcAlias -> stagingAlias
destAlias -> targetAlias

Tang8330 (approved):

clients/bigquery/dialect/dialect.go
nathan-artie: Sure.

PR #666, opened 2024-05-15
[snowflake] Add tests for building additional equality strings

nathan-artie:


clients/snowflake/writes.go
```diff
@@ -49,16 +60,8 @@ func (s *Store) Merge(tableData *optimization.TableData) error {
 			}
 		}
 
-		var additionalEqualityStrings []string
```
nathan-artie, line 52: Shouldn't be any changes here, just moved to a method.

Tang8330 (approved):

PR #670, opened 2024-05-15
[sql] Change staging table alias to `stg`

nathan-artie:


clients/bigquery/dialect/dialect.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	stagingAlias = "cc"
```
Tang8330, line 19: Can we move this to a shared constant vs. pkg level?

Tang8330 (approved):

clients/bigquery/dialect/dialect.go
nathan-artie: Yep.

PR #677, opened 2024-05-15
[DWH] Update Append Interface

Tang8330:
The previous PR: #676 added the ability to exclude deleted columns, but it did not add it to the `.Append(...)` interface.


Tang8330:
Closing this PR because it overcomplicates things and doesn't have good separation of concerns.

PR #690, opened 2024-05-17
[BigQuery] Missed one spot for aliasing

Tang8330:
As per title

### DIFF
--- a/clients/bigquery/merge.go
+++ b/clients/bigquery/merge.go
@@ -2,7 +2,6 @@ package bigquery
 
 import (
 	"fmt"
-	"log/slog"
 
 	"cloud.google.com/go/bigquery"
 
@@ -42,8 +41,7 @@ func (s *Store) Merge(tableData *optimization.TableData) error {
 
 		mergeString, err := generateMergeString(tableData.TopicConfig().BigQueryPartitionSettings, s.Dialect(), distinctDates)
 		if err != nil {
-			slog.Warn("Failed to generate merge string", slog.Any("err", err))
-			return err
+			return fmt.Errorf("failed to generate merge string: %w", err)
 		}
 
 		additionalEqualityStrings = []string{mergeString}
--- a/clients/bigquery/merge.go
+++ b/clients/bigquery/merge.go
@@ -4,21 +4,17 @@ import (
 	"fmt"
 	"log/slog"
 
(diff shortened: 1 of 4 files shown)


nathan-artie (approved):

lib/kafkalib/partition/settings.go
```diff
@@ -43,7 +44,7 @@ func (b *BigQuerySettings) GenerateMergeString(values []string) (string, error)
 	case "time":
 		switch b.PartitionBy {
 		case "daily":
-			return fmt.Sprintf(`DATE(c.%s) IN (%s)`, b.PartitionField, array.StringsJoinAddSingleQuotes(values)), nil
+			return fmt.Sprintf(`DATE(%s.%s) IN (%s)`, constants.TargetAlias, b.PartitionField, array.StringsJoinAddSingleQuotes(values)), nil
```
nathan-artie, line 47: Sneaky!

nathan-artie: Is `b.PartitionField` always a column? Should we be quoting it?

nathan-artie: If so could do `QuoteTableAliasColumns(constants.TargetAlias, columns.Column(b.PartitionField, typing.Invalid))`

nathan-artie: I think using the `sql` package at all may cause an import cycle, so maybe something to look at after this PR.

Tang8330: Yes, we should escape it

Tang8330: I'll leave a TODO

Tang8330: Actually, I can just move this to the bq pkg

clients/bigquery/merge.go
```diff
@@ -49,3 +53,29 @@ func (s *Store) Merge(tableData *optimization.TableData) error {
 		RetryColBackfill: true,
 	})
 }
+
+func generateMergeString(bqSettings *partition.BigQuerySettings, dialect sql.Dialect, values []string) (string, error) {
```
nathan-artie, line 57: Nice.

nathan-artie (approved):

PR #695, opened 2024-05-18
Replace `array.Empty` with `stringutil.Empty`

nathan-artie:
We have two of the same function.

### DIFF
--- a/lib/array/strings.go
+++ b/lib/array/strings.go
@@ -52,15 +52,3 @@ func InterfaceToArrayString(val any, recastAsArray bool) ([]string, error) {
 
 	return vals, nil
 }
-
-// Empty will iterate over a list, if one of the item in the list is empty, it will return true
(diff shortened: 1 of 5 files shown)


lib/array/strings.go
```diff
@@ -52,15 +52,3 @@ func InterfaceToArrayString(val any, recastAsArray bool) ([]string, error) {
 
 	return vals, nil
 }
-
-// Empty will iterate over a list, if one of the item in the list is empty, it will return true
-// This is useful to check the presence of a setting.
-func Empty(list []string) bool {
```
nathan-artie, line 58: Compare with https://github.com/artie-labs/transfer/blob/9a35dced4f2965172333e763e6ae9dadb666d0a9/lib/stringutil/strings.go#L32

Tang8330 (approved):

PR #697, opened 2024-05-21
Improve BigQuery `castColVal`

Tang8330:
This is needed because Postgres and MongoDB return JSON objects differently.

* Mongo returns it with the native type e.g.: `map[string]any{}`
* Postgres and others has already JSON marshalled the data by the time we receive it

As a result, `castColVal(...)` must handle structs that have been marshalled already and ones that have not.

### DIFF
--- a/clients/bigquery/cast.go
+++ b/clients/bigquery/cast.go
@@ -56,6 +56,7 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
 			return extTime.String(dialect.BQStreamingTimeFormat), nil
 		}
 	case typing.Struct.Kind:
+		// TODO: See if we can improve this eval and find a better location, see: https://github.com/artie-labs/transfer/pull/697#discussion_r1609280164
 		if strings.Contains(fmt.Sprint(colVal), constants.ToastUnavailableValuePlaceholder) {
 			return fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder), nil
 		}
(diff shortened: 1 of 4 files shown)


clients/bigquery/cast.go
```diff
@@ -59,11 +60,23 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
 			return fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder), nil
 		}
 
-		colValString, isOk := colVal.(string)
```
nathan-artie, line 62: Would it make sense to move the check above for the `constants.ToastUnavailableValuePlaceholder` to the `if` statement below? Otherwise we're potentially calling `fmt.Sprint()` on a map.

clients/bigquery/cast.go
```diff
@@ -59,11 +60,23 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
...
-			// Empty string is not a valid JSON object, so let's return nil.
-			return nil, nil
+		// Structs from relational and Mongo are different.
+		// MongoDB will return the native objects back such as `map[string]any{"hello": "world"}`
+		// Relational will return a string representation of the struct such as `{"hello": "world"}`
+		if colValString, isOk := colVal.(string); isOk {
+			if colValString == "" {
+				return nil, nil
+			}
+
```
nathan-artie, line 70: Would it make sense to move the check above for the `constants.ToastUnavailableValuePlaceholder` here? Otherwise we're potentially calling `fmt.Sprint()` on a map.

Tang8330: Even if we did call `fmt.Sprint(...)` on a map, it'd be fine? We're really just checking to make sure the value doesn't contain `__debezium_unavailable_value`

nathan-artie: Yeah it'd still work, I was mostly thinking that in the case that `colVal` is a large nested data structure it would be inefficient to convert it to a string.

Tang8330:
Can leave a TODO to have a more elegant solution? This whole function
should be improved since we are converting a bunch of data types to a
string.

nathan-artie (approved):

PR #699, opened 2024-05-24
[BigQuery] Improve castColVal function

Tang8330:
We were converting most values to a string and letting the BigQuery API deal with all the type conversions.

This PR makes it so that we stop doing that for ints, floats and booleans. In addition, I also rewrote the tests to be more succinct

### DIFF
--- a/clients/bigquery/cast.go
+++ b/clients/bigquery/cast.go
@@ -3,6 +3,7 @@ package bigquery
 import (
 	"encoding/json"
 	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/artie-labs/transfer/clients/bigquery/dialect"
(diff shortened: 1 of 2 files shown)


clients/bigquery/cast.go
```diff
@@ -18,8 +18,9 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
 	if colVal == nil {
 		return nil, nil
 	}
-
 	switch colKind.KindDetails.Kind {
+	case typing.Float.Kind, typing.Integer.Kind, typing.Boolean.Kind:
```
nathan-artie, line 22: If we add an explicit case for `typing.String` we can then make the default case throw an error that the column kind is not supported which will catch potential errors if we add more typing kinds.

nathan-artie (approved):

PR #700, opened 2024-05-25
Improve handling default values

Tang8330:
## Changes

1. Simplifying ShouldParseComments
2. Get Table config to be cognizant of default values for Microsoft SQL Server

This is part one of supporting default values for SQL Server, we still need to figure out a way to write the default values constraint here: https://github.com/artie-labs/transfer/blob/2b8727c50eaa5a2cf573068722b09b69347165fc/clients/shared/utils.go#L62


clients/shared/table_config.go
```diff
@@ -27,38 +31,16 @@ type GetTableCfgArgs struct {
...
-	tableConfig := g.ConfigMap.TableConfig(g.TableID)
-	if tableConfig != nil {
+	if tableConfig := g.ConfigMap.TableConfig(g.TableID); tableConfig != nil {
 		return tableConfig, nil
 	}
 
-	rows, err := g.Dwh.Query(g.Query, g.Args...)
-	defer func() {
-		if rows != nil {
-			err = rows.Close()
```
Tang8330, line 54: From Golang:

```go
// Close closes the [Rows], preventing further enumeration. If [Rows.Next] is called
// and returns false and there are no further result sets,
// the [Rows] are closed automatically and it will suffice to check the
// result of [Rows.Err]. Close is idempotent and does not affect the result of [Rows.Err].
```

clients/mssql/store.go
```diff
@@ -75,11 +74,10 @@ func (s *Store) Dedupe(_ sql.TableIdentifier, _ []string, _ kafkalib.TopicConfig
 }
 
 func (s *Store) GetTableConfig(tableData *optimization.TableData) (*types.DwhTableConfig, error) {
-	// TODO: Figure out how to leave a comment.
 	const (
 		describeNameCol        = "column_name"
 		describeTypeCol        = "data_type"
-		describeDescriptionCol = "description"
+		describeDescriptionCol = "column_default"
```
nathan-artie, line 80: Is this using `ColumnDescLabel` field as the default value for MSSQL and the column description for other databases types? Can we use different fields so we don't confuse the two?

Tang8330: Yes it is. I'll rename

Tang8330: Err, slight correction after reading this for the second time.

We're querying this field to see what the default value is.

clients/shared/table_config.go
```diff
@@ -15,50 +15,30 @@ import (
...
 
-func (g GetTableCfgArgs) ShouldParseComment(comment string) bool {
-	if g.EmptyCommentValue != nil && comment == *g.EmptyCommentValue {
-		return false
-	}
-
-	// Snowflake and Redshift both will return `<nil>` if the comment does not exist, this will check the value.
-	// BigQuery returns ""
-	return true
+	ColumnNameForName        string
```
Tang8330, line 29: @nathan-artie I renamed these to be more clear

clients/shared/table_config.go
```diff
@@ -15,50 +15,31 @@ import (
 	"github.com/artie-labs/transfer/lib/typing/columns"
 )
 
+func isCommentSet(comment string) bool {
+	return !(comment == "" || comment == "<nil>")
```
nathan-artie, line 19: Where is `<nil>` coming from? Are we coercing a null into a string somewhere where we could just change it to an empty string?

clients/shared/table_config.go
nathan-artie: For MSSQL though `ColumnNameForDescription` is misleading because it's not the column name used for the description like the other data sources, it's the column name that stores the default value. Unless you're thinking of the default value as a type of description?

I also find it somewhat confusing later on when we do:
```
if comment, isOk := row[g.ColumnNameForDescription]; isOk && isCommentSet(comment) {
``` 
Because the code leads me to believe that this column is storing a comment, but for MSSQL we're storing the default value not a comment.

clients/shared/table_config.go
Tang8330: `<nil>` is a side effect of https://github.com/artie-labs/transfer/blob/master/clients/shared/table_config.go#L102

clients/shared/table_config.go
Tang8330: For OLAP sources, we're relying on the comment column to know whether this column has been backfilled.

OLTP sources have a concept of default values, which we're getting the data from as you can see from the SQL query

clients/shared/table_config.go
nathan-artie: Can we make that code just write an empty string instead?

clients/shared/table_config.go
nathan-artie: I would say we should just have two different fields, a `ColumnNameForComment` and a `ColumnNameForDefaultValue`.

clients/shared/table_config.go
Tang8330: Sure, I'll leave a TODO so we can do it after default values are handled

clients/shared/table_config.go
Tang8330: Sure

clients/shared/table_config.go
nathan-artie: Or change it to `row := make(map[string]*string)`.

nathan-artie (approved):

clients/shared/table_config.go
```diff
@@ -102,25 +82,28 @@ func (g GetTableCfgArgs) GetTableConfig() (*types.DwhTableConfig, error) {
...
 
-		col := columns.NewColumn(row[g.ColumnNameLabel], kindDetails)
-		comment, isOk := row[g.ColumnDescLabel]
-		if isOk && g.ShouldParseComment(comment) {
+		col := columns.NewColumn(row[g.ColumnNameForName], kindDetails)
+		if _, isOk := row[g.ColumnNameForDefaultValue]; isOk && isCommentSet(row[g.ColumnNameForDefaultValue]) {
+			col.SetBackfilled(true)
+		}
+
+		if comment, isOk := row[g.ColumnNameForComment]; isOk && isCommentSet(comment) {
```
nathan-artie, line 99: Should this be in an `else` block so that it doesn't run if col.SetBackfilled(true) has been set?

PR #701, opened 2024-05-28
[Typing] Improve the way we parse MongoDB

Tang8330:
## Context

The way we currently handle parsing MongoDB results has slight inefficiencies and causes us to not be able to delineate between floats and ints.

The current way is:
1. Take JSON extended string
2. Parse that into BSON
3. Run our types registry to convert BSON objects into regular values
4. JSON marshal into `map[string]any`

By running JSON marshal, we are losing the ability to delineate between floats and ints.

## Change

This PR directly takes the BSON object and converts it into `map[string]any` natively, without JSON marshal.

This change reduces the time it takes for us to parse by 27% (see below)

### Before
```bash
BenchmarkJSONEToMap-12    	  122318	    199390 ns/op
```

### After
```bash
BenchmarkJSONEToMap-12    	  166398	    145455 ns/op
```

### DIFF
--- a/lib/typing/mongo/bson.go
+++ b/lib/typing/mongo/bson.go
@@ -83,7 +83,7 @@ func bsonBinaryValueToMap(value primitive.Binary) (any, error) {
 func bsonValueToGoValue(value any) (any, error) {
 	switch v := value.(type) {
 	case primitive.DateTime:
-		return time.Unix(0, int64(v)*int64(time.Millisecond)).UTC().Format(ext.ISO8601), nil
+		return v.Time().UTC().Format(ext.ISO8601), nil
 	case primitive.ObjectID:
 		return v.Hex(), nil
 	case primitive.Binary:
--- a/lib/typing/mongo/bson.go
+++ b/lib/typing/mongo/bson.go
@@ -34,11 +34,9 @@ func JSONEToMap(val []byte) (map[string]any, error) {
 		return nil, fmt.Errorf("failed to unmarshal ext json: %w", err)
 	}
 
-	// Directly decode bsonDoc to map[string]any
 	return bsonDocToMap(bsonDoc)
 }
 
-// Helper function to convert bson.D to map[string]any
 func bsonDocToMap(doc bson.D) (map[string]any, error) {
 	result := make(map[string]any)
 	for _, elem := range doc {
@@ -76,14 +74,14 @@ func bsonBinaryValueToMap(value primitive.Binary) (any, error) {
 		}
 
 		return parsedUUID.String(), nil
+	default:
+		return map[string]any{
+			"$binary": map[string]any{
+				"base64":  base64.StdEncoding.EncodeToString(value.Data),
+				"subType": fmt.Sprintf("%02x", value.Subtype),
+			},
+		}, nil
 	}
-
-	return map[string]any{
-		"$binary": map[string]any{
-			"base64":  base64.StdEncoding.EncodeToString(value.Data),
-			"subType": fmt.Sprintf("%02x", value.Subtype),
-		},
-	}, nil
 }
 
 // bsonValueToGoValue - https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/
--- a/lib/typing/mongo/bson.go
+++ b/lib/typing/mongo/bson.go
@@ -93,7 +93,18 @@ func bsonValueToGoValue(value any) (any, error) {
 	case primitive.Binary:
 		return bsonBinaryValueToMap(v)
 	case primitive.Decimal128:
-		return strconv.ParseFloat(v.String(), 64)
+		potentialFloat, err := strconv.ParseFloat(v.String(), 64)
+		if err != nil {
+			return nil, err
+		}
+
+		if v.String() == fmt.Sprint(potentialFloat) {
+			return potentialFloat, nil
+		}
+
+		// If the string representation of the Decimal128 is not equal to the float representation
+		// then we return the string representation
+		return v.String(), nil
 	case primitive.Timestamp:
 		return time.Unix(int64(v.T), 0).UTC().Format(ext.ISO8601), nil
 	case bson.D:
--- a/lib/typing/mongo/bson.go
+++ b/lib/typing/mongo/bson.go
@@ -89,6 +89,7 @@ func bsonValueToGoValue(value any) (any, error) {
 	case primitive.Binary:
 		return bsonBinaryValueToMap(v)
 	case primitive.Decimal128:
+		// We purposefully chose a string representation here because not all systems can correctly handle Decimal128 without losing precision
 		return v.String(), nil
 	case primitive.Timestamp:
 		return time.Unix(int64(v.T), 0).UTC().Format(ext.ISO8601), nil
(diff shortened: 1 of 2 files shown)


Tang8330:
Write a benchmark file

lib/typing/mongo/bson.go
```diff
@@ -32,133 +27,101 @@ func JSONEToMap(val []byte) (map[string]any, error) {
...
-	}
+func bsonValueToGoValue(value any) (any, error) {
+	switch v := value.(type) {
+	case primitive.DateTime:
+		return time.Unix(0, int64(v)*int64(time.Millisecond)).UTC().Format(ext.ISO8601), nil
+	case primitive.ObjectID:
+		return v.Hex(), nil
+	case primitive.Binary:
+		return bsonBinaryValueToMap(v)
+	case primitive.Decimal128:
```
nathan-artie, line 92: Do we need to handle decimals that between 64 and 128 bits in size?

Tang8330: wdym?

Tang8330: If it exceeds float64, we'll return it as a string

lib/typing/mongo/bson.go
```diff
@@ -32,133 +27,101 @@ func JSONEToMap(val []byte) (map[string]any, error) {
...
 
-func createCustomRegistry() *bsoncodec.Registry {
-	rb := bson.NewRegistry()
-	rb.RegisterTypeEncoder(tDateTime, bsoncodec.ValueEncoderFunc(dateTimeEncodeValue))
-	rb.RegisterTypeEncoder(tOID, bsoncodec.ValueEncoderFunc(objectIDEncodeValue))
-	rb.RegisterTypeEncoder(tBinary, bsoncodec.ValueEncoderFunc(binaryEncodeValue))
-	rb.RegisterTypeEncoder(tDecimal, bsoncodec.ValueEncoderFunc(decimalEncodeValue))
-	rb.RegisterTypeEncoder(tTimestamp, bsoncodec.ValueEncoderFunc(timestampEncodeValue))
-	return rb
+		// If the string representation of the Decimal128 is not equal to the float representation
```
nathan-artie, line 102: In what cases would the string representation be different? Is this something we expect to happen often or something we should log if it happens?

Tang8330: On a practical level, float64 should be sufficient. However, in theory a decimal128 has more precision than float64.

So it is possible that a number becomes looses precision with f64

Tang8330: https://en.wikipedia.org/wiki/Decimal128_floating-point_format 

https://en.wikipedia.org/wiki/Decimal64_floating-point_format

lib/typing/mongo/bson.go
```diff
@@ -32,133 +27,101 @@ func JSONEToMap(val []byte) (map[string]any, error) {
...
+	case bson.D:
+		return bsonDocToMap(v)
+	case bson.A:
+		return bsonArrayToSlice(v)
+	case primitive.MaxKey:
+		return map[string]any{"$maxKey": 1}, nil
+	case primitive.MinKey:
+		return map[string]any{"$minKey": 1}, nil
+	case primitive.JavaScript:
+		return map[string]any{"$code": fmt.Sprint(v)}, nil
```
nathan-artie, line 116: Can probably do `string(v)` since `primitive.JavaScript` is just a `type JavaScript string`.

lib/typing/mongo/bson.go
```diff
@@ -32,133 +27,101 @@ func JSONEToMap(val []byte) (map[string]any, error) {
...
+		return bsonBinaryValueToMap(v)
+	case primitive.Decimal128:
+		potentialFloat, err := strconv.ParseFloat(v.String(), 64)
+		if err != nil {
+			return nil, err
+		}
 
-	return vw.WriteString(time.Unix(int64(s.T), 0).UTC().Format(ext.ISO8601))
-}
+		if v.String() == fmt.Sprint(potentialFloat) {
```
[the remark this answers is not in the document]
Tang8330: Nice, this lowered the time by 190 ns

lib/typing/mongo/bson.go
nathan-artie: Won't we get a `value out of range` for numbers too big to be represented by float64, like `1.7E+309`?

Tang8330: I expected to - but we didn't.

See: https://github.com/artie-labs/transfer/blob/15cabd47fa783ce5e7a0b967d399857e5b6aa26e/lib/typing/mongo/bson_test.go#L216-L221

lib/typing/mongo/bson.go
nathan-artie: I see, is there a disadvantage to just always returning a string then? I feel like it might be harder to work with these values if sometimes they are float64s and sometimes they are strings.

Another thing I'm wondering is whether we'll end up with false positives for relatively small numbers such as `1234567.0` since when they are run through `fmt.Sprintf()` they will be output in exponential notation (e.g. `1.234567e+06`) and then will be different from what `Decimal128.String()` outputs.

Tang8330: I'm not opposed to the idea of outputting strings to be more deterministic. BSON also has `int64/float64` that folks can use

Tang8330: Convert to string then?

lib/typing/mongo/bson.go
nathan-artie: I think that's just because the `fmt.Sprint()` is truncating the value to `1.7976931348623157e+308` when it converts to exponential notation and that value is representable by a float64. If you change it to `primitive.ParseDecimal128("1.7e+309")` you do get an out of range error.

lib/typing/mongo/bson_test.go
```diff
@@ -173,3 +179,88 @@ func TestMarshal(t *testing.T) {
...
+func TestBsonDocToMap(t *testing.T) {
+	doc := bson.D{
+		{"foo", "bar"},
+	}
+	result, err := bsonDocToMap(doc)
+	assert.NoError(t, err)
+	assert.Equal(t, map[string]any{"foo": "bar"}, result)
+}
+
+func TestBsonValueToGoVale(t *testing.T) {
```
nathan-artie, line 192: Nit: `bsonValueToGoVale` is missing a "u".

lib/typing/mongo/bson.go
```diff
@@ -32,133 +26,91 @@ func JSONEToMap(val []byte) (map[string]any, error) {
...
-		return bsoncodec.ValueEncoderError{Name: "ObjectIDEncodeValue", Types: []reflect.Type{tBinary}, Received: val}
-	}
-
-	s, isOk := val.Interface().(primitive.Timestamp)
-	if !isOk {
-		return bsoncodec.ValueEncoderError{Name: "ObjectIDEncodeValue not Binary", Types: []reflect.Type{tBinary}, Received: val}
+func bsonValueToGoValue(value any) (any, error) {
+	switch v := value.(type) {
+	case primitive.DateTime:
+		return time.Unix(0, int64(v)*int64(time.Millisecond)).UTC().Format(ext.ISO8601), nil
```
nathan-artie, line 86: Can you do: `v.Time().UTC().Format(ext.ISO8601)`?

Tang8330: Yup I can

nathan-artie (approved):

PR #708, opened 2024-06-02
[DateTime] Solve Edge Case

Tang8330:
## Problem

There's an edge case right now with DateTimeWithTimeZone

Consider this, when Debezium gives us the field, it'll look like this:
```json
{
    "type": "string",
    "optional": false,
    "name": "io.debezium.time.ZonedTimestamp",
    "version": 1,
    "field": "due_date"
}
```

As a result, the string can look something like this: 
```
+275760-09-13T00:00:00.000000Z
```

This will fail the Go `time.Time` parsing logic. However, this is technically a validate datetime in Postgres.

### DIFF
--- a/lib/debezium/types.go
+++ b/lib/debezium/types.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"log/slog"
+	"strings"
 	"time"
 
 	"github.com/artie-labs/transfer/lib/config/constants"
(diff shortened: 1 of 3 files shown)


lib/debezium/types.go
```diff
@@ -130,6 +131,27 @@ func (f Field) ParseValue(value any) (any, error) {
...
+		}
+
+		// We don't need to pass `additionalDateFormats` because this data type layout is standardized by Debezium
+		extTime, err := ext.ParseExtendedDateTime(dtString, nil)
+		if err == nil {
+			return extTime, nil
+		}
+
+		// Check if the year exceeds 9999, or is negative
+		if parts := strings.Split(dtString, "-"); len(parts) == 3 {
```
nathan-artie, line 147: If the year is negative there will be four parts, the first being an empty string.

Tang8330: True, added a guardrail above

nathan-artie (approved):

PR #720, opened 2024-06-13
[bigquery] Add scaffolding for the Storage Write API

nathan-artie:
Breaking out some parts of the Storage Write API PR I'm working out.

### DIFF
--- a/clients/bigquery/cast.go
+++ b/clients/bigquery/cast.go
@@ -110,7 +110,9 @@ func EncodeStructToJSONString(value any) (string, error) {
 
 	stringValue := string(bytes)
 	if strings.Contains(stringValue, constants.ToastUnavailableValuePlaceholder) {
+		// TODO: Remove this if we don't see it in the logs.
 		slog.Error("encoded JSON value contains the toast unavailable value placeholder")
+		return fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder), nil
 	}
 	return stringValue, nil
 }
--- a/clients/bigquery/cast.go
+++ b/clients/bigquery/cast.go
@@ -96,16 +96,21 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
 // MongoDB will return the native objects back such as `map[string]any{"hello": "world"}`
 // Relational will return a string representation of the struct such as `{"hello": "world"}`
 func EncodeStructToJSONString(value any) (string, error) {
-	if colValString, isOk := value.(string); isOk {
-		if strings.Contains(colValString, constants.ToastUnavailableValuePlaceholder) {
+	if stringValue, isOk := value.(string); isOk {
+		if strings.Contains(stringValue, constants.ToastUnavailableValuePlaceholder) {
 			return fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder), nil
 		}
-		return colValString, nil
+		return stringValue, nil
 	}
 
-	colValBytes, err := json.Marshal(value)
+	bytes, err := json.Marshal(value)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


clients/bigquery/cast.go
```diff
@@ -104,3 +90,22 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
 	slog.Error("Unexpected BigQuery Data Type", slog.Any("colKind", colKind.KindDetails.Kind), slog.Any("colVal", colVal))
 	return fmt.Sprint(colVal), nil
 }
+
+// EncodeStructToJSONString takes a struct as either a string or Go object and encodes it into a JSON string.
+// Structs from relational and Mongo are different.
+// MongoDB will return the native objects back such as `map[string]any{"hello": "world"}`
+// Relational will return a string representation of the struct such as `{"hello": "world"}`
+func EncodeStructToJSONString(value any) (string, error) {
```
nathan-artie, line 98: Will need this for the Storage Write API.

clients/bigquery/cast.go
```diff
@@ -104,3 +90,22 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
...
+	if colValString, isOk := value.(string); isOk {
+		if strings.Contains(colValString, constants.ToastUnavailableValuePlaceholder) {
+			return fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder), nil
+		}
+		return colValString, nil
+	}
+
+	colValBytes, err := json.Marshal(value)
```
nathan-artie, line 106: @Tang8330 if `value` isn't a string do we care if the toast placeholder appears within it (e.g. `["__debezium_unavailable_value"]`, `{"foo": "__debezium_unavailable_value"}`)?

lib/config/bigquery.go
```diff
@@ -5,11 +5,12 @@ import "fmt"
...
-	DefaultDataset    string `yaml:"defaultDataset"`
-	ProjectID         string `yaml:"projectID"`
-	Location          string `yaml:"location"`
-	BatchSize         int    `yaml:"batchSize"`
+	PathToCredentials  string `yaml:"pathToCredentials"`
+	DefaultDataset     string `yaml:"defaultDataset"`
+	ProjectID          string `yaml:"projectID"`
+	Location           string `yaml:"location"`
+	BatchSize          int    `yaml:"batchSize"`
+	UseStorageWriteAPI bool   `yaml:"__useStorageWriteAPI"` // Not officially supported yet.
```
nathan-artie, line 13: For config settings that are not intended for production use I think it's nice to prefix them with "__" to indicate that the are not officially supported, that way we can get rid of them without breaking backwards compatibility and having to bump the version.

clients/bigquery/cast.go
Tang8330: I don't think it's possible, but can we add a guardrail for it and Sentry when it does happen?

lib/config/bigquery.go
Tang8330: Sure

clients/bigquery/cast.go
```diff
@@ -96,16 +96,21 @@ func castColVal(colVal any, colKind columns.Column, additionalDateFmts []string)
...
-	colValBytes, err := json.Marshal(value)
+	bytes, err := json.Marshal(value)
 	if err != nil {
 		return "", fmt.Errorf("failed to marshal colVal: %w", err)
 	}
-	return string(colValBytes), nil
+
+	stringValue := string(bytes)
+	if strings.Contains(stringValue, constants.ToastUnavailableValuePlaceholder) {
+		slog.Error("encoded JSON value contains the toast unavailable value placeholder")
```
Tang8330, line 113: This should return this too

```go
return fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder), nil
```

PR #726, opened 2024-06-17
[BigQuery] Use StorageWrite API + Dedupe

Tang8330:
## Problem

The BigQuery streaming API (InsertAll or Storage Write) do not guarantee exactly once semantics. This means that we'll need to de-duplicate the staging tables before we merge them into the target table.

This PR does exactly that.

### DIFF
--- a/clients/bigquery/dialect/dialect.go
+++ b/clients/bigquery/dialect/dialect.go
@@ -123,7 +123,7 @@ func (BigQueryDialect) IsColumnAlreadyExistsErr(err error) bool {
 	return strings.Contains(err.Error(), "Column already exists")
 }
 
-func (BigQueryDialect) IsTableDoesNotExistErr(err error) bool {
+func (BigQueryDialect) IsTableDoesNotExistErr(_ error) bool {
 	return false
 }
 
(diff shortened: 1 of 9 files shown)


clients/bigquery/dialect/dialect.go
```diff
@@ -154,6 +154,15 @@ func (bd BigQueryDialect) BuildIsNotToastValueExpression(tableAlias constants.Ta
 	return fmt.Sprintf("COALESCE(%s != '%s', true)", colName, constants.ToastUnavailableValuePlaceholder)
 }
 
+func (bd BigQueryDialect) GetDedupeTableQuery(tableID sql.TableIdentifier, primaryKeys []string) string {
+	primaryKeysEscaped := sql.QuoteIdentifiers(primaryKeys, bd)
+	return fmt.Sprintf(`(SELECT * FROM %s QUALIFY ROW_NUMBER() OVER (PARTITION BY %s ORDER BY %s) = 1)`,
```
Tang8330, line 159: Had to do this because BigQuery doesn't like DISTINCTs for JSON columns, see:

```
Column foo of type JSON cannot be used in SELECT DISTINCT
```

clients/bigquery/dialect/dialect.go
```diff
@@ -154,6 +154,15 @@ func (bd BigQueryDialect) BuildIsNotToastValueExpression(tableAlias constants.Ta
 	return fmt.Sprintf("COALESCE(%s != '%s', true)", colName, constants.ToastUnavailableValuePlaceholder)
 }
 
+func (bd BigQueryDialect) GetDedupeTableQuery(tableID sql.TableIdentifier, primaryKeys []string) string {
```
nathan-artie, line 157: Nit: Call this `BuildDedupeTableQuery` so it matches the rest of the query methods on the dialect.

nathan-artie (approved):

clients/bigquery/dialect/dialect.go
nathan-artie: Want to add a comment to this effect?

Tang8330: Didn't see this. Yeah I'll add one

PR #729, opened 2024-06-17
[BigQuery] Removing Legacy API

Tang8330:
With: https://github.com/artie-labs/transfer/pull/726 we are now using Storage Write API. This PR is removing the legacy Insert API


nathan-artie (approved):
🔪 🔪 🔪

PR #730, opened 2024-06-17
Refactor dedupe signature

Tang8330:
We only need `includeArtieUpdatedAt`, we don't need to pass in the whole `kafkalib.TopicConfig`


nathan-artie (approved):
Nice!

PR #734, opened 2024-06-17
[Decimal] Fix Encoding Problem

Tang8330:
Our previous library was susceptible to precision loss when converting from a `big.Float` into a `big.Int`. Repro: https://go.dev/play/p/qg8NODRlxi5

### DIFF
--- a/lib/debezium/decimal.go
+++ b/lib/debezium/decimal.go
@@ -2,6 +2,7 @@ package debezium
 
 import (
 	"fmt"
+	"math"
 	"math/big"
 
 	"github.com/artie-labs/transfer/lib/typing/decimal"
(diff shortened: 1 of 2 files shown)


lib/debezium/decimal.go
```diff
@@ -9,15 +9,20 @@ import (
 
 // EncodeDecimal is used to encode a string representation of a number to `org.apache.kafka.connect.data.Decimal`.
 func EncodeDecimal(value string, scale int) ([]byte, error) {
```
nathan-artie, line 11: We should probably make `scale` an `int64`.

Tang8330: Really? I think if we were to change it, it should be like `uint16`

Tang8330: > PostgreSQL permits the scale in a numeric type declaration to be any value in the range -1000 to 1000. However, the SQL standard requires the scale to be in the range 0 to precision. Using scales outside that range may not be portable to other database systems.

Tang8330: Oh, are you mentioning this because we then don't need to convert on L17?

```go
big.NewInt(int64(scale))
```

nathan-artie (approved):

lib/debezium/decimal.go
nathan-artie: I just think that when it comes to mathematical operations we should avoid `int` and always try to use an integer with an explicit amount of bits (`int16`, `int32`, `int64`, etc). I would be fine with `int16` here.

Tang8330: Yeah sgtm

Tang8330: I'm gonna add a TODO since it'll require refactoring our decimal type too.

PR #735, opened 2024-06-17
Add `batch.BySize` function

nathan-artie:


lib/batch/batch.go
```diff
@@ -0,0 +1,48 @@
+package batch
+
+import "fmt"
+
+// BySize takes a series of elements, encodes them, groups them into batches of bytes that sum to at most [maxSize], and
+// then passes each of those batches to the [yield] function.
+func BySize[T any](in []T, maxSize int, encode func(T) ([]byte, error), yield func([][]byte) error) error {
```
Tang8330, line 7: maxSizeBytes?

lib/batch/batch_test.go
```diff
@@ -0,0 +1,90 @@
...
+	{
+		// Non-empty slice + all items exactly fit into one batch:
+		batches, err := testBySize([]string{"foo", "bar", "baz", "qux"}, 12, goodEncoder)
+		assert.NoError(t, err)
+		assert.Len(t, batches, 1)
+		assert.Equal(t, [][]byte{[]byte("foo"), []byte("bar"), []byte("baz"), []byte("qux")}, batches[0])
+	}
+	{
+		// Non-empty slice + all items exactly fit into just under one batch:
+		batches, err := testBySize([]string{"foo", "bar", "baz", "qux"}, 13, goodEncoder)
```
Tang8330, line 85: Worth adding a check on how many times `goodEncoding` was called? Which will test how many times `yield` was called

Tang8330 (approved):

lib/batch/batch_test.go
nathan-artie: In this case `goodEncoder` will be called once for each item (4 times) but `yield` will be called once. The amount of times that yield is called will always equal the length of `batches`.

danafallon (approved):

PR #741, opened 2024-06-18
[bigquery] Increase Storage Write API max request size

nathan-artie:
Also kill `batchSize` config.

### DIFF
--- a/clients/bigquery/bigquery.go
+++ b/clients/bigquery/bigquery.go
@@ -34,8 +34,8 @@ const (
 	describeNameCol               = "column_name"
 	describeTypeCol               = "data_type"
 	describeCommentCol            = "description"
-	// Storage Write API is limited to 10 MB, let's start out conservative and use 80% of that.
-	maxRequestByteSize = 10_000_000 * .8
+	// Storage Write API is limited to 10 MiB, subtract 50 KiB to account for request overhead.
+	maxRequestByteSize = (10 * 1024 * 1024) - (50 * 1024)
 )
 
 type Store struct {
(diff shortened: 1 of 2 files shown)


clients/bigquery/bigquery.go
```diff
@@ -34,13 +34,12 @@ const (
 	describeNameCol               = "column_name"
 	describeTypeCol               = "data_type"
 	describeCommentCol            = "description"
-	// Storage Write API is limited to 10 MB, let's start out conservative and use 80% of that.
-	maxRequestByteSize = 10_000_000 * .8
+	// Storage Write API is limited to 10 MiB, subtract 50 KiB to account for request overhead.
+	maxRequestByteSize = (10 * 1024 * 1024) - (50 * 1024)
```
Tang8330, line 38: is 50 kb fixed?

clients/bigquery/bigquery.go
```diff
@@ -223,7 +222,6 @@ func (s *Store) Dedupe(tableID sql.TableIdentifier, primaryKeys []string, includ
 }
 
 func LoadBigQuery(cfg config.Config, _store *db.Store) (*Store, error) {
-	cfg.BigQuery.LoadDefaultValues()
```
Tang8330, line 226: Nice

Tang8330 (approved):

clients/bigquery/bigquery.go
nathan-artie: After a bunch of testing the cutoff seemed to be around 30 KiB so I decided to go with 50 KiB to be safe.

Tang8330:
i c

clients/bigquery/bigquery.go
nathan-artie: I tested it it with a pretty wide table so I think it should be fine, but if we do see errors we can increase it to 100 KiB.

PR #749, opened 2024-06-22
Use uint16 for `EncodeDecimal`

Tang8330:
As per title.

### DIFF
--- a/lib/debezium/decimal_test.go
+++ b/lib/debezium/decimal_test.go
@@ -10,7 +10,7 @@ func TestEncodeDecimal(t *testing.T) {
 	testCases := []struct {
 		name  string
 		value string
-		scale int
+		scale uint16
 
 		expectedErr string
 	}{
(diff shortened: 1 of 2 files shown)


lib/debezium/decimal_test.go
```diff
@@ -86,7 +86,7 @@ func TestEncodeDecimal(t *testing.T) {
 	for _, testCase := range testCases {
 		encodedValue, err := EncodeDecimal(testCase.value, testCase.scale)
 		if testCase.expectedErr == "" {
-			decodedValue := DecodeDecimal(encodedValue, nil, testCase.scale)
+			decodedValue := DecodeDecimal(encodedValue, nil, int(testCase.scale))
```
nathan-artie, line 89: It'd be nice to make make `DecodeDecimal` and `Decimal` use `int16` too.

nathan-artie (approved):

lib/debezium/decimal_test.go
Tang8330: Yeah I tried, it's pretty hard

nathan-artie: What about int64 then?

Tang8330:
The precision isn't the problem, it's the fact that we have lots of code
references that will need to specify

PR #752, opened 2024-06-24
[go] Fix bug with encoding negative decimal numbers

nathan-artie:
When encoding a negative decimal number we have to convert the bytes into two's complement. In two's complement the first bit is reserved for the sign, however for certain numbers the sign is already a positive, so in that case we need to pad the output by one empty byte so that the first bit can safely be set to 0 without changing the value. Specifically we were encoding values `-65,281` through `-65,535` incorrectly.

### DIFF
--- a/lib/debezium/decimal.go
+++ b/lib/debezium/decimal.go
@@ -29,7 +29,8 @@ func EncodeDecimal(value string, scale uint16) ([]byte, error) {
 		// Convert to two's complement if the number is negative
 
 		if data[0] > 127 {
-			// Prepend an empty byte if the first bit is already set.
+			// If the first bit is already set then it is a significant bit and we need to prepend an additional byte
+			// that will be used to indicate whether the number is positive or negative.
 			data = slices.Concat([]byte{0}, data)
 		}
 
--- a/lib/debezium/decimal.go
+++ b/lib/debezium/decimal.go
@@ -31,7 +31,7 @@ func EncodeDecimal(value string, scale uint16) ([]byte, error) {
 		if data[0] >= 0x80 {
 			// If the first bit is already set then it is a significant bit and we need to prepend an additional byte
 			// so that the first bit can safely be used to indicate whether the number is positive or negative.
-			data = slices.Concat([]byte{0}, data)
+			data = slices.Concat([]byte{0x00}, data)
 		}
 
 		// Inverting bits for two's complement.
--- a/lib/debezium/decimal.go
+++ b/lib/debezium/decimal.go
@@ -30,7 +30,7 @@ func EncodeDecimal(value string, scale uint16) ([]byte, error) {
 
 		if data[0] > 127 {
 			// If the first bit is already set then it is a significant bit and we need to prepend an additional byte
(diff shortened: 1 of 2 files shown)


lib/debezium/decimal.go
```diff
@@ -23,11 +24,14 @@ func EncodeDecimal(value string, scale uint16) ([]byte, error) {
 		return nil, fmt.Errorf("unable to use %q as a floating-point number", value)
 	}
 
-	data := bigIntValue.Bytes()
+	data := bigIntValue.Bytes() // [Bytes] returns the absolute value of the number.
 	if bigIntValue.Sign() < 0 {
 		// Convert to two's complement if the number is negative
-		bigIntValue = bigIntValue.Neg(bigIntValue)
```
nathan-artie, line 29: No need to do this as the `Bytes()` call already uses the absolute value of the number.

lib/debezium/decimal_test.go
```diff
@@ -71,6 +88,16 @@ func TestEncodeDecimal(t *testing.T) {
 			value: "1.05",
 			scale: 2,
 		},
+		{
+			name:  "negative number: 2^16 - 255",
+			value: "-65281",
```
nathan-artie, line 93: Without this fix `-65281` becomes `-1`.

lib/debezium/decimal_test.go
```diff
@@ -71,6 +88,16 @@ func TestEncodeDecimal(t *testing.T) {
...
+			scale: 0,
+		},
+		{
+			name:  "negative number: 2^16 - 1",
+			value: "-65535",
```
nathan-artie, line 98: Without this fix `-65535` becomes `-255`.

lib/debezium/decimal_test.go
```diff
@@ -84,10 +120,9 @@ func TestEncodeDecimal(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
-		encodedValue, err := EncodeDecimal(testCase.value, testCase.scale)
+		actual, err := encodeDecode(testCase.value, testCase.scale)
 		if testCase.expectedErr == "" {
-			decodedValue := DecodeDecimal(encodedValue, nil, int(testCase.scale))
-			assert.Equal(t, testCase.value, decodedValue.String(), testCase.name)
+			assert.Equal(t, testCase.value, actual, testCase.name)
```
Tang8330, line 125: One thing I realized, we're not actually checking for the err

Tang8330 (approved):

lib/debezium/decimal_test.go
nathan-artie: Ooh yeah good call!

lib/debezium/decimal.go
```diff
@@ -23,11 +24,14 @@ func EncodeDecimal(value string, scale uint16) ([]byte, error) {
...
-		data = bigIntValue.Bytes()
+
+		if data[0] > 127 {
+			// Prepend an empty byte if the first bit is already set.
+			data = slices.Concat([]byte{0}, data)
```
Tang8330, line 33: Why do we need to prepend an empty byte?

PR #756, opened 2024-06-25
Simplify `debezium.encodeBigInt` - pt. 2

nathan-artie:


lib/debezium/decimal.go
```diff
@@ -34,12 +34,7 @@ func encodeBigInt(value *big.Int) []byte {
 		twoComplement := new(big.Int).SetBytes(data)
 		twoComplement.Add(twoComplement, big.NewInt(1))
 
-		data = twoComplement.Bytes()
-		if data[0]&0x80 == 0 {
```
nathan-artie, line 38: This code only executes when the first bit is not set, but it will always be true now so we don't need this code. I verified by checking all numbers in the range (`-10,000,000`, `10,000,000`).

Tang8330 (approved):

PR #760, opened 2024-06-25
[debezium] Fix bug with `EncodeDecimal`

nathan-artie:
While doing some testing I discovered that if we encode and decode `-9063701308.217222135` with `EncodeDecimal`/`DecodeDecimal` the result comes back as `-9063701308.217222134`. In order to fix this and mitigate other potential issues with decimals this switches us to use https://github.com/cockroachdb/apd for decimals.

### DIFF
--- a/lib/debezium/decimal_test.go
+++ b/lib/debezium/decimal_test.go
@@ -82,6 +82,7 @@ func TestDecimalWithNewExponent(t *testing.T) {
 	assert.Equal(t, "12.34", decimalWithNewExponent(mustParseDecimalFromString("12.349"), -2).Text('f'))
 	assert.Equal(t, "12.3", decimalWithNewExponent(mustParseDecimalFromString("12.349"), -1).Text('f'))
 	assert.Equal(t, "12", decimalWithNewExponent(mustParseDecimalFromString("12.349"), 0).Text('f'))
+	assert.Equal(t, "10", decimalWithNewExponent(mustParseDecimalFromString("12.349"), 1).Text('f'))
 }
 
 func TestEncodeDecimal(t *testing.T) {
--- a/lib/debezium/decimal_test.go
+++ b/lib/debezium/decimal_test.go
@@ -195,6 +195,16 @@ func TestEncodeDecimal(t *testing.T) {
 			scale: 0,
 		},
 		{
+			name:  "number with a scale of 15",
+			value: "0.000022998904125",
+			scale: 15,
+		},
+		{
+			name:  "number with a scale of 15",
+			value: "145.183000000000000",
+			scale: 15,
+		},
+		{
 			name:        "malformed - empty string",
 			value:       "",
 			expectedErr: `unable to use "" as a floating-point number`,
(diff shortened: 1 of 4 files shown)


lib/debezium/decimal_test.go
```diff
@@ -67,6 +94,32 @@ func TestEncodeDecimal(t *testing.T) {
...
+	assert.Equal(t, "145.1830000000000010", mustEncodeAndDecodeDecimal("145.183000000000001", 16))
+	assert.Equal(t, "-145.1830000000000010", mustEncodeAndDecodeDecimal("-145.183000000000001", 16))
+	assert.Equal(t, "145.1830000000000040", mustEncodeAndDecodeDecimal("145.183000000000004", 16))
+	assert.Equal(t, "-145.1830000000000040", mustEncodeAndDecodeDecimal("-145.183000000000004", 16))
+	assert.Equal(t, "145.1830000000000050", mustEncodeAndDecodeDecimal("145.183000000000005", 16))
+	assert.Equal(t, "-145.1830000000000050", mustEncodeAndDecodeDecimal("-145.183000000000005", 16))
+	assert.Equal(t, "145.1830000000000090", mustEncodeAndDecodeDecimal("145.183000000000009", 16))
+	assert.Equal(t, "-145.1830000000000090", mustEncodeAndDecodeDecimal("-145.183000000000009", 16))
+
+	assert.Equal(t, "-9063701308.217222135", mustEncodeAndDecodeDecimal("-9063701308.217222135", 9))
```
nathan-artie, line 121: This line was failing before.

lib/debezium/decimal_test.go
```diff
@@ -56,6 +57,34 @@ func mustEncodeAndDecodeDecimal(value string, scale uint16) string {
...
+	assert.Equal(t, "00", decimalWithNewExponent(apd.New(0, 0), 1).Text('f'))
+	assert.Equal(t, "0.0", decimalWithNewExponent(apd.New(0, 0), -1).Text('f'))
+
+	// Same exponent:
+	assert.Equal(t, "12.349", decimalWithNewExponent(mustParseDecimalFromString("12.349"), -3).Text('f'))
+	// More precise exponent:
+	assert.Equal(t, "12.3490", decimalWithNewExponent(mustParseDecimalFromString("12.349"), -4).Text('f'))
+	assert.Equal(t, "12.34900", decimalWithNewExponent(mustParseDecimalFromString("12.349"), -5).Text('f'))
+	// Lest precise exponent:
+	// Extra digits should be truncated rather than rounded.
```
danafallon, line 81: Should we add a test case where the first truncated digit is >= 5 to make sure it will be truncated instead of rounded up?

nathan-artie: Isn't that the next line:
> assert.Equal(t, "12.34", decimalWithNewExponent(mustParseDecimalFromString("12.349"), -2).Text('f'))

danafallon: Oh yeah you're right, I missed that it was truncating before the 9. Carry on!

PR #761, opened 2024-06-25
[debezium] Change function signature of `EncodeDecimal`

nathan-artie:
`EncodeDecimal` is only called by Reader in two places, and in one of those places we already have code to determine the scale, which is something we can simplify if we have a `apd.Decimal`, so rather than depend on Transfer to parse strings into encoded `org.apache.kafka.connect.data.Decimal`s we'll let Reader handle the parsing to `apd.Decimal` and just let Transfer handle the encoding.

### DIFF
--- a/lib/debezium/decimal.go
+++ b/lib/debezium/decimal.go
@@ -100,8 +100,8 @@ func EncodeDecimal(decimal *apd.Decimal) ([]byte, int32) {
 	return encodeBigInt(bigIntValue), -decimal.Exponent
 }
 
-// EncodeDecimalWithScale is used to encode a [apd.Decimal] to [org.apache.kafka.connect.data.Decimal] using a specific
-// scale.
+// EncodeDecimalWithScale is used to encode a [apd.Decimal] to [org.apache.kafka.connect.data.Decimal]
+// using a specific scale.
 func EncodeDecimalWithScale(decimal *apd.Decimal, scale int32) []byte {
 	targetExponent := -scale // Negate scale since [Decimal.Exponent] is negative.
 	if decimal.Exponent != targetExponent {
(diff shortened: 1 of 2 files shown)


lib/debezium/decimal.go
```diff
@@ -90,24 +89,26 @@ func decimalWithNewExponent(decimal *apd.Decimal, newExponent int32) *apd.Decima
...
-	}
-
-	targetExponent := -int32(scale) // Negate scale since [Decimal.Exponent] is negative.
-	if decimal.Exponent != targetExponent {
-		decimal = decimalWithNewExponent(decimal, targetExponent)
-	}
-
+// EncodeDecimal is used to encode a [apd.Decimal] to [org.apache.kafka.connect.data.Decimal].
+// The scale of the value (which is the negated exponent of the decimal) is returned as the second argument.
+func EncodeDecimal(decimal *apd.Decimal) ([]byte, int32) {
```
nathan-artie, line 94: I chose to use `int32` for the scale because that's both what `apd.Decimal` uses as well as the wire representation of `org.apache.kafka.connect.data.Decimal`.

Tang8330: Make this private?

Tang8330 (approved):

lib/debezium/decimal.go
Tang8330: Unless you are planning to have reader call EncodeDecimal and EncodeDecimalWithScale depending on whether we know the scale or not

nathan-artie: Reader is going to call both functions in different places.

nathan-artie: Hah your second comment didn't show up for me when I responded!

PR #762, opened 2024-06-25
When soft deleting, preserve previous values if we have them in memory

danafallon:
Step 1 of not clearing out the values of a row when it's soft deleted.

If a row is updated and then deleted before the next flush, we have its previous values in memory so we can use those and just add the deleted flag.

### DIFF
--- a/lib/optimization/table_data.go
+++ b/lib/optimization/table_data.go
@@ -132,11 +132,8 @@ func (t *TableData) InsertRow(pk string, rowData map[string]any, delete bool) {
 	if isOk {
 		prevRowSize = size.GetApproxSize(prevRow)
 		if delete {
-			// If the row was deleted, preserve the previous values if we have them in memory
+			// If the row was deleted, preserve the previous values that we have in memory
 			rowData = prevRow
-			if len(rowData) == 0 {
-				rowData = map[string]any{}
-			}
 			rowData[constants.DeleteColumnMarker] = true
 		} else {
 			for key, val := range rowData {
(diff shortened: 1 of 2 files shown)


lib/optimization/table_data.go
```diff
@@ -131,17 +131,27 @@ func (t *TableData) InsertRow(pk string, rowData map[string]any, delete bool) {
 	prevRow, isOk := t.rowsData[pk]
 	if isOk {
```
danafallon, line 132: (hide whitespace to make the diff simpler)

lib/optimization/table_data.go
```diff
@@ -131,17 +131,27 @@ func (t *TableData) InsertRow(pk string, rowData map[string]any, delete bool) {
...
-		for key, val := range rowData {
-			if val == constants.ToastUnavailableValuePlaceholder {
-				// Copy it from prevRow.
-				prevVal, isOk := prevRow[key]
-				if !isOk {
-					continue
+		if delete {
+			// If the row was deleted, preserve the previous values if we have them in memory
+			rowData = prevRow
+			if len(rowData) == 0 {
```
Tang8330, line 137: How would we have a len(rowData) == 0 ?

Tang8330: I'm pretty sure the only way this would happen is if we had a [tombstone message](https://docs.google.com/document/d/1ApCs6EewoIlUUHkJGDAJLOs7f13AyB2AhuAAk020FJM/edit), but we're skipping those now

danafallon: A [test was panicking](https://github.com/artie-labs/transfer/actions/runs/9670133429/job/26678253059#step:5:53) because it was inserting a row that was a nil map. Initially I made the test provide an empty map instead of nil, but then I thought it would be safer to handle that here so we won't get a panic if a nil row does get inserted for any reason.

Tang8330: Hm got it. Do you want to throw an error instead?

Tang8330: Or get rid of the test?

Tang8330 (approved):

PR #765, opened 2024-06-26
[debezium] Return `apd.Decimal` from `DecodeDecimal`

nathan-artie:
Because `DecodeDecimal` is using `big.Float` there are small discrepancies with decoding certain numbers. For example, decoding the binary representation of `-74961544796695.89960242` produces `-74961544796695.89960241`.


lib/debezium/decimal.go
```diff
@@ -62,33 +62,6 @@ func decodeBigInt(data []byte) *big.Int {
 	return bigInt
 }
 
-// decimalWithNewExponent takes a [apd.Decimal] and returns a new [apd.Decimal] with a the given exponent.
-// If the new exponent is less precise then the extra digits will be truncated.
-func decimalWithNewExponent(decimal *apd.Decimal, newExponent int32) *apd.Decimal {
```
nathan-artie, line 67: Moved this to https://github.com/artie-labs/transfer/pull/765/files#diff-05c5ea56a4927faef9a74d525d6312544af5a191a532c2d8cc2d9fd4ee59ecefR113 to get around circular imports, no changes.

lib/typing/decimal/decimal.go
```diff
@@ -92,3 +107,30 @@ func (d *Decimal) BigQueryKind() string {
 
 	return "STRING"
 }
+
+// DecimalWithNewExponent takes a [apd.Decimal] and returns a new [apd.Decimal] with a the given exponent.
```
nathan-artie, line 111: No changes to this function.

lib/debezium/decimal_test.go
```diff
@@ -132,6 +112,7 @@ func TestEncodeDecimalWithScale(t *testing.T) {
 	assert.Equal(t, "-145.1830000000000090", mustEncodeAndDecodeDecimal("-145.183000000000009", 16))
 
 	assert.Equal(t, "-9063701308.217222135", mustEncodeAndDecodeDecimal("-9063701308.217222135", 9))
+	assert.Equal(t, "-74961544796695.89960242", mustEncodeAndDecodeDecimal("-74961544796695.89960242", 8))
```
nathan-artie, line 115: This test fails on master.

clients/shared/utils.go
```diff
@@ -44,7 +44,7 @@ func DefaultValue(column columns.Column, dialect sql.Dialect, additionalDateFmts
 			return nil, fmt.Errorf("colVal is not type *decimal.Decimal")
 		}
 
-		return val.Value(), nil
+		return val.String(), nil
```
nathan-artie, line 47: @Tang8330  This is the only (non-test) place we call `Decimal.Value()` I'm wondering if we can just call `.String()` instead?

lib/typing/decimal/decimal.go
```diff
@@ -52,17 +52,15 @@ func (d *Decimal) Precision() *int {
...
+	targetExponent := -int32(d.scale)
+	value := d.value
+	if value.Exponent != targetExponent {
+		value = DecimalWithNewExponent(value, targetExponent)
 	}
+	return value.Text('f')
+}
 
-	// Depending on the precision, we will want to convert value to STRING or keep as a FLOAT.
+func (d *Decimal) Value() *apd.Decimal {
```
nathan-artie, line 63: We were only calling `.Value()` in one place so I changed it to call `.String()` instead.

lib/debezium/types_bench_test.go
```diff
@@ -19,7 +19,7 @@ func BenchmarkDecodeDecimal_P64_S10(b *testing.B) {
 		assert.NoError(b, err)
 		dec, err := field.DecodeDecimal(bytes)
 		assert.NoError(b, err)
-		assert.Equal(b, "123456789012345678901234567890123456789012345678901234.1234567889", dec.Value())
+		assert.Equal(b, "123456789012345678901234567890123456789012345678901234.1234567890", dec.String())
```
nathan-artie, line 22: I think this was decoding incorrectly.

lib/debezium/decimal_test.go
```diff
@@ -49,32 +49,14 @@ func mustParseDecimal(value string) *apd.Decimal {
 	return decimal
 }
 
-func TestDecimalWithNewExponent(t *testing.T) {
```
nathan-artie, line 52: Moved this.

clients/shared/utils.go
nathan-artie: Especially since `.Value()` used to return a string when there was no precision or the precision was too high.

clients/bigquery/storagewrite_test.go
```diff
@@ -129,9 +129,9 @@ func TestRowToMessage(t *testing.T) {
 		"c_float_int32":    int32(1234),
 		"c_float_int64":    int64(1234),
 		"c_float_string":   "4444.55555",
-		"c_numeric":        decimal.NewDecimal(nil, 5, big.NewFloat(3.1415926)),
+		"c_numeric":        decimal.NewDecimal(nil, numbers.MustParseDecimal("3.14159")),
```
nathan-artie, line 132: We now rely on the exponent of the decimal instead of passing in a scale, so I the decimal places here were trimmed to 5.

lib/debezium/types_bench_test.go
nathan-artie, line 22: This was decoding correctly, the unscaled value is `1234567890123456789012345678901234567890123456789012341234567890`

nathan-artie: Which makes sense since this number is just `1234567890` repeating (with a 1234 before the decimal point).

danafallon (approved):

PR #767, opened 2024-06-26
[typing] Split out `DecimalDetails`

nathan-artie:
This way `decimal.Decimal` is just used for storing decimal values and not details about numeric columns.

### DIFF
--- a/lib/typing/decimal/details.go
+++ b/lib/typing/decimal/details.go
@@ -1,0 +1,64 @@
+package decimal
+
+import (
+	"fmt"
+
+	"github.com/artie-labs/transfer/lib/ptr"
+)
+
+type DecimalDetails struct {
+	scale     int
+	precision *int
+}
+
+func NewDecimalDetails(precision *int, scale int) *DecimalDetails {
+	if precision != nil {
+		if scale > *precision && *precision != -1 {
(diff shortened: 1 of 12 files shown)


lib/typing/decimal/details.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/artie-labs/transfer/lib/ptr"
+)
+
+type DecimalDetails struct {
+	scale     int
+	precision *int
+}
+
+func NewDecimalDetails(precision *int, scale int) *DecimalDetails {
```
nathan-artie, line 14: This is exactly the same as `NewDecimal()` except that it doesn't take a third `value` argument.

lib/typing/decimal/details.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (d DecimalDetails) Scale() int {
+	return d.scale
+}
+
+func (d DecimalDetails) Precision() *int {
+	return d.precision
+}
+
+// SnowflakeKind - is used to determine whether a NUMERIC data type should be a STRING or NUMERIC(p, s).
```
nathan-artie, line 39: Move the following methods without changes.

Tang8330 (approved):
Sweet!

PR #771, opened 2024-06-26
[typing] Remove pointer for `DecimalDetails` precision

nathan-artie:
Instead of having two ways to indicate no precision set (`nil` and `-1`) we'll use `-1`.

### DIFF
--- a/lib/debezium/schema.go
+++ b/lib/debezium/schema.go
@@ -93,9 +93,14 @@ func (f Field) ToKindDetails() typing.KindDetails {
 	case JSON, GeometryPointType, GeometryType, GeographyType:
 		return typing.Struct
 	case KafkaDecimalType:
-		scale, precision, err := f.GetScaleAndPrecision()
+		scale, precisionPtr, err := f.GetScaleAndPrecision()
 		if err != nil {
 			return typing.Invalid
+		}
+
+		var precision int32 = decimal.PrecisionNotSpecified
+		if precisionPtr != nil {
+			precision = *precisionPtr
 		}
 
 		eDecimal := typing.EDecimal
@@ -106,7 +111,7 @@ func (f Field) ToKindDetails() typing.KindDetails {
 		// This is because scale is not specified at the column level, rather at the row level
 		// It shouldn't matter much anyway since the column type we are creating is `TEXT` to avoid boundary errors.
 		eDecimal := typing.EDecimal
-		eDecimal.ExtendedDecimalDetails = decimal.NewDecimalDetails(ptr.ToInt32(decimal.PrecisionNotSpecified), decimal.DefaultScale)
+		eDecimal.ExtendedDecimalDetails = decimal.NewDecimalDetails(decimal.PrecisionNotSpecified, decimal.DefaultScale)
 		return eDecimal
 	}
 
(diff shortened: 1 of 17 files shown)


lib/debezium/schema.go
```diff
@@ -93,11 +93,16 @@ func (f Field) ToKindDetails() typing.KindDetails {
 	case JSON, GeometryPointType, GeometryType, GeographyType:
 		return typing.Struct
 	case KafkaDecimalType:
-		scale, precision, err := f.GetScaleAndPrecision()
+		scale, precisionPtr, err := f.GetScaleAndPrecision()
 		if err != nil {
 			return typing.Invalid
 		}
 
+		var precision int32 = decimal.PrecisionNotSpecified
```
Tang8330, line 101: precision := decimal.PrecisionNotSpecified

lib/typing/decimal/decimal.go
```diff
@@ -72,5 +72,9 @@ func (d *Decimal) Value() any {
 }
 
 func (d *Decimal) Details() DecimalDetails {
-	return DecimalDetails{scale: d.Scale(), precision: d.precision}
+	var precision int32 = PrecisionNotSpecified
```
Tang8330, line 75: precision := PrecisionNotSpecified

lib/typing/decimal/base.go
```diff
@@ -31,18 +31,13 @@ func (d *DecimalDetails) isBigNumeric() bool {
 	}
 
 	// max(1,s) <= p <= s + 38
-	return numbers.BetweenEq(max(1, d.scale), d.scale+38, *d.precision)
+	return numbers.BetweenEq(max(1, d.scale), d.scale+38, d.precision)
 }
 
 func (d *DecimalDetails) toKind(maxPrecision int32, exceededKind string) string {
-	precision := maxPrecision
```
nathan-artie, line 38: There is a subtle change here in that before if `d.precision` was nil we would use the maxPrecision and now we'll treat it the same as `-1`.

lib/typing/decimal/details.go
```diff
@@ -2,24 +2,20 @@ package decimal
...
-	precision *int32
+	precision int32
 }
 
-func NewDecimalDetails(precision *int32, scale int32) *DecimalDetails {
-	if precision != nil {
-		if scale > *precision && *precision != -1 {
-			// Note: -1 precision means it's not specified.
+func NewDecimalDetails(precision int32, scale int32) *DecimalDetails {
+	if scale > precision && precision != -1 {
```
Tang8330, line 13: Use constant `PrecisionNotSpecified` instead of -1

lib/typing/parquet.go
```diff
@@ -113,7 +113,7 @@ func (k *KindDetails) ParquetAnnotation(colName string) (*Field, error) {
 		}, nil
 	case EDecimal.Kind:
 		precision := k.ExtendedDecimalDetails.Precision()
-		if precision == nil || *precision == -1 {
+		if precision == -1 {
```
Tang8330, line 116: Use constant

lib/typing/decimal/base.go
Tang8330: I think that's fine

Tang8330 (approved):

PR #772, opened 2024-06-26
[typing] Remove pointer from `Decimal` precision

nathan-artie:


lib/typing/decimal/decimal.go
```diff
@@ -37,14 +34,22 @@ func NewDecimal(precision *int32, value *apd.Decimal) *Decimal {
...
 func (d *Decimal) Scale() int32 {
 	return -d.value.Exponent
 }
 
-func (d *Decimal) Precision() *int32 {
+func (d *Decimal) Precision() int32 {
 	return d.precision
 }
 
+func (d *Decimal) Value() *apd.Decimal {
```
nathan-artie, line 49: Added this in case we want to inspect the value.

Tang8330 (approved):

PR #773, opened 2024-06-27
[typing] Kill `Decimal.Value` method

nathan-artie:


clients/shared/utils.go
```diff
@@ -44,8 +44,7 @@ func DefaultValue(column columns.Column, dialect sql.Dialect, additionalDateFmts
 			return nil, fmt.Errorf("colVal is not type *decimal.Decimal")
 		}
 
-		// TODO: Call [String] instead.
-		return val.Value(), nil
+		return val.String(), nil
```
nathan-artie, line 47: This is the only place it was used.

nathan-artie: And this function is only called when backfilling.

nathan-artie: @Tang8330 I feel like since we were sometimes emitting a string before, always emitting a string should be fine?

Tang8330: Yeah it's fine, I just tested it

Tang8330 (approved):

clients/shared/utils.go
nathan-artie: Sweet thanks!

PR #774, opened 2024-06-27
Rename `shared/utils` -> `shared/default_value`

Tang8330:
No code change


clients/shared/default_value.go
danafallon, line 1: Should `BackfillColumn` go in a separate file?

Tang8330: Backfill column is used as part of updating default values

danafallon (approved):

PR #776, opened 2024-06-27
Decimal clean up.

Tang8330:
## Changes

1. Renaming DecimalDetails -> Details
2. Removing pointers since it's unnecessary

### DIFF
--- a/lib/typing/decimal/details.go
+++ b/lib/typing/decimal/details.go
@@ -9,7 +9,7 @@ type Details struct {
 	precision int32
 }
 
-func NewDecimalDetails(precision int32, scale int32) *Details {
+func NewDetails(precision int32, scale int32) Details {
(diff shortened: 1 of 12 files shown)


lib/typing/decimal/details.go
```diff
@@ -4,12 +4,12 @@ import (
...
 )
 
-type DecimalDetails struct {
+type Details struct {
 	scale     int32
 	precision int32
 }
 
-func NewDecimalDetails(precision int32, scale int32) *DecimalDetails {
+func NewDecimalDetails(precision int32, scale int32) *Details {
```
nathan-artie, line 12: Probably don't need a pointer here either.

nathan-artie: And we should rename it to `NewDetails`.

nathan-artie (approved):

PR #778, opened 2024-06-27
[typing] Clean up `NewDecimalWithPrecision`

nathan-artie:
We don't need to duplicate the logic for handling invalid decimal precisions since we never read from `Decimal.Precision()`.


lib/typing/decimal/details.go
```diff
@@ -4,6 +4,13 @@ import (
 	"fmt"
 )
 
+const (
+	DefaultScale int32 = 5
```
nathan-artie, line 8: Moved these constants too since they are only used for decimal details.

lib/typing/decimal/decimal.go
```diff
@@ -4,30 +4,15 @@ import (
...
-)
-
 func NewDecimalWithPrecision(value *apd.Decimal, precision int32) *Decimal {
-	scale := -value.Exponent
-	if scale > precision && precision != PrecisionNotSpecified {
-		// Note: -1 precision means it's not specified.
-
-		// This is typically not possible, but Postgres has a design flaw that allows you to do things like: NUMERIC(5, 6) which actually equates to NUMERIC(7, 6)
-		// We are setting precision to be scale + 1 to account for the leading zero for decimal numbers.
-		precision = scale + 1
```
Tang8330, line 28: How come we don't need this?

nathan-artie: Because we have this exact logic in `NewDetails()` and we don't read from `Decimal.Precision()` except for tests.

Tang8330: Got it. We should refactor the precision function then

```
func (d *Decimal) Precision() int32 {
	return NewDetails(d.precision, d.Scale()).Precision()
}
```

nathan-artie: I'd just as soon as kill `Precision`, we don't need it.

Tang8330 (approved):

PR #779, opened 2024-06-27
[Redshift] Addressing KindDetails TODO

Tang8330:
## Changes

1. Hard fail if string precision fails to parse
2. Hard fail if the data type is not recognized


clients/redshift/dialect/dialect.go
```diff
@@ -75,15 +74,14 @@ func (RedshiftDialect) KindForDataType(rawType string, stringPrecision string) (
...
-		if err == nil {
-			strPrecision = &precision
+		if err != nil {
+			return typing.Invalid, fmt.Errorf("failed to parse string precision: %q, err: %w", stringPrecision, err)
 		}
 
 		return typing.KindDetails{
 			Kind:                    typing.String.Kind,
-			OptionalStringPrecision: strPrecision,
+			OptionalStringPrecision: &precision,
```
nathan-artie, line 84: It would be nice to make `OptionalStringPrecision` an `int32` and use ParseInt instead of `Atoi`.

nathan-artie (approved):

clients/redshift/dialect/dialect.go
Tang8330: I'll leave a TODO there. `OptionalStringPrecision` touches a lot of different parts of our code

Tang8330:
Pushed a new image `v1.25.33-exp`, will deploy this and let it run in Production for a couple of days before I merge.

PR #782, opened 2024-06-28
Upgrade Alpine image to 3.20

nathan-artie:


Dockerfile
```diff
@@ -1,14 +1,13 @@
-FROM --platform=linux/amd64 golang:1.22-alpine AS builder
+FROM --platform=linux/amd64 golang:1.22-alpine3.20 AS builder
```
Tang8330, line 1: Is this file still being used?

Tang8330 (approved):

Dockerfile
nathan-artie: Hmm, I'm guessing it's not since we have a dockerfile just for goreleaser and that's what we use for builds.

PR #786, opened 2024-07-06
Deprecate `createAllColumnsIfAvailable` setting

Tang8330:
If the schema is available, let's return the type even if the value is null. This is the more desirable behavior and we have been setting this to true on our managed deployments for some time now.

### DIFF
--- a/lib/typing/parse.go
+++ b/lib/typing/parse.go
@@ -25,11 +25,6 @@ func ParseValue(settings Settings, key string, optionalSchema map[string]KindDet
 		}
 	}
 
-	if val == nil {
-		// Skip parsing if the value is null and schema is not available.
-		return Invalid
-	}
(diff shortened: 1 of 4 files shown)


lib/typing/parse.go
```diff
@@ -30,6 +25,11 @@ func ParseValue(settings Settings, key string, optionalSchema map[string]KindDet
 		}
 	}
 
+	if val == nil {
```
nathan-artie, line 28: Since `parseValue` is going to return `Invalid` for a `nil` value do we need this check?

nathan-artie (approved):

lib/typing/parse.go
Tang8330: Good point, let's remove.

PR #787, opened 2024-07-09
Add `ENTRYPOINT` to Dockerfile

nathan-artie:


goreleaser.dockerfile
```diff
@@ -1,2 +1,3 @@
 FROM --platform=linux/amd64 alpine:3.20
 COPY transfer /transfer
+ENTRYPOINT ["/transfer"]
```
Tang8330, line 3: How's this going to work if we need to pass in a config file via the CLI?

nathan-artie: The commands you enter in the CLI are appended to the `ENTRYPOINT`.

Tang8330 (approved):

PR #789, opened 2024-07-09
[Kafka] Refactor the way we handle Kafka connections

Tang8330:
This PR creates a `Connection` struct that can be turned into a Dialer and Writer. 

This will remove the boilerplate in our dashboard and Reader.

### DIFF
--- a/lib/kafkalib/connection.go
+++ b/lib/kafkalib/connection.go
@@ -84,3 +84,38 @@ func (c Connection) Dialer(ctx context.Context) (*kafka.Dialer, error) {
 
 	return dialer, nil
 }
+
+func (c Connection) Transport() (*kafka.Transport, error) {
+	transport := &kafka.Transport{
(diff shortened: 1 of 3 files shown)


lib/kafkalib/connection.go
```diff
@@ -0,0 +1,119 @@
...
+	case Plain:
+		// No mechanism
+	default:
+		return nil, fmt.Errorf("unsupported kafka mechanism: %s", c.Mechanism())
+	}
+
+	return dialer, nil
+}
+
+func (c Connection) Transport() (*kafka.Transport, error) {
```
danafallon, line 86: Is this used anywhere yet?

danafallon (approved):
Just one question, but looks good

lib/kafkalib/connection.go
Tang8330: No, I want to use it here though: https://github.com/artie-labs/reader/pull/434

PR #793, opened 2024-07-16
Clean up more snowflake dialect tests

danafallon:
Following https://github.com/artie-labs/transfer/pull/792, this removes the subquery from the other snowflake test cases that used one.

I also added a helper function to build the columns in a deterministic order so each test case doesn't have to implement that separately.


Tang8330 (approved):
Nice

PR #796, opened 2024-07-17
Snowflake: don't null out column values when soft-deleting

danafallon:
When a row is soft-deleted, now we'll preserve the previous values of all the columns (other than the `__artie_delete` column) instead of overwriting them all to be empty.

- Added a temporary column marker to indicate if a row should have only its `__artie_delete` column updated
- Made snowflake's merge function check that value to decide how to merge
- Made all dialects filter out this temporary column before merging so it doesn't end up in the destination

As a followup, I'll make the other dialects use the new marker when merging.

### DIFF
--- a/clients/bigquery/dialect/dialect.go
+++ b/clients/bigquery/dialect/dialect.go
@@ -249,9 +249,9 @@ MERGE INTO %s %s USING %s AS %s ON %s`,
 		tableID.FullyQualifiedName(), constants.TargetAlias, subQuery, constants.StagingAlias, strings.Join(equalitySQLParts, " AND "),
 	)
 
-	cols, removed := columns.RemoveOnlySetDeletedColumnMarker(cols)
+	cols, removed := columns.RemoveOnlySetDeleteColumnMarker(cols)
 	if !removed {
-		return []string{}, errors.New("artie only_set_deleted flag doesn't exist")
+		return []string{}, errors.New("artie only_set_delete flag doesn't exist")
 	}
 
 	if softDelete {
--- a/clients/bigquery/dialect/dialect.go
+++ b/clients/bigquery/dialect/dialect.go
@@ -249,6 +249,10 @@ MERGE INTO %s %s USING %s AS %s ON %s`,
 		tableID.FullyQualifiedName(), constants.TargetAlias, subQuery, constants.StagingAlias, strings.Join(equalitySQLParts, " AND "),
 	)
 
+	cols, removed := columns.RemoveOnlySetDeletedColumnMarker(cols)
+	if !removed {
+		return []string{}, errors.New("artie set deleted only flag doesn't exist")
+	}
 	if softDelete {
 		return []string{baseQuery + fmt.Sprintf(`
 WHEN MATCHED %sTHEN UPDATE SET %s
@@ -263,7 +267,7 @@ WHEN NOT MATCHED THEN INSERT (%s) VALUES (%s);`,
 	}
 
 	// We also need to remove __artie flags since it does not exist in the destination table
-	cols, removed := columns.RemoveDeleteColumnMarker(cols)
+	cols, removed = columns.RemoveDeleteColumnMarker(cols)
 	if !removed {
 		return []string{}, errors.New("artie delete flag doesn't exist")
 	}
--- a/clients/bigquery/dialect/dialect.go
+++ b/clients/bigquery/dialect/dialect.go
@@ -251,8 +251,9 @@ MERGE INTO %s %s USING %s AS %s ON %s`,
 
 	cols, removed := columns.RemoveOnlySetDeletedColumnMarker(cols)
 	if !removed {
-		return []string{}, errors.New("artie set deleted only flag doesn't exist")
+		return []string{}, errors.New("artie only_set_deleted flag doesn't exist")
 	}
+
 	if softDelete {
 		return []string{baseQuery + fmt.Sprintf(`
 WHEN MATCHED %sTHEN UPDATE SET %s
--- a/clients/bigquery/dialect/dialect.go
+++ b/clients/bigquery/dialect/dialect.go
@@ -255,6 +255,7 @@ MERGE INTO %s %s USING %s AS %s ON %s`,
 	}
 
 	if softDelete {
+		// TODO alter this merge query to update only the __artie_delete column if OnlySetDeleteColumnMarker is true
 		return []string{baseQuery + fmt.Sprintf(`
 WHEN MATCHED %sTHEN UPDATE SET %s
 WHEN NOT MATCHED THEN INSERT (%s) VALUES (%s);`,
--- a/clients/mssql/dialect/dialect.go
+++ b/clients/mssql/dialect/dialect.go
@@ -205,6 +205,7 @@ USING %s AS %s ON %s`,
 	}
 
 	if softDelete {
+		// TODO alter this merge query to update only the __artie_delete column if OnlySetDeleteColumnMarker is true
 		return []string{baseQuery + fmt.Sprintf(`
 WHEN MATCHED %sTHEN UPDATE SET %s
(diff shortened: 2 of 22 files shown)


lib/cdc/mongo/debezium.go
```diff
@@ -162,6 +162,10 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]any, tc *kafkalib.TopicCon
 		}
 
 		retMap[constants.DeleteColumnMarker] = true
+		// For now, assume we only want to set the deleted column and leave other values alone.
+		// If previous values for the other columns are in memory (not flushed yet), [TableData.InsertRow] will handle
+		// filling them in and setting this to false.
+		retMap[constants.OnlySetDeleteColumnMarker] = true
```
danafallon, line 168: This is where we first set this marker; we set it to true if it's a delete event, false otherwise

lib/cdc/util/relational_event.go
```diff
@@ -92,6 +92,10 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]any, tc *kafkalib.TopicCon
 		// We _can_ rely on *before* since even without running replicate identity, it will still copy over
 		// the PK. We can explore simplifying this interface in the future by leveraging before.
 		retMap[constants.DeleteColumnMarker] = true
+		// For now, assume we only want to set the deleted column and leave other values alone.
+		// If previous values for the other columns are in memory (not flushed yet), [TableData.InsertRow] will handle
+		// filling them in and setting this to false.
+		retMap[constants.OnlySetDeleteColumnMarker] = true
```
danafallon, line 98: Same as https://github.com/artie-labs/transfer/pull/796/files#r1681761970

lib/optimization/table_data.go
```diff
@@ -131,8 +131,13 @@ func (t *TableData) InsertRow(pk string, rowData map[string]any, delete bool) {
 	if prevRow, isOk := t.rowsData[pk]; isOk {
 		prevRowSize = size.GetApproxSize(prevRow)
 		if delete {
-			// If the row was deleted, preserve the previous values that we have in memory
+			// If the row was deleted, preserve the previous values that we have in memory,
+			// unless the previous row was also a delete operation
 			rowData = prevRow
+			prevRowDeleted, isOk := prevRow[constants.DeleteColumnMarker]
+			if !isOk || !prevRowDeleted.(bool) {
+				rowData[constants.OnlySetDeleteColumnMarker] = false
```
danafallon, line 139: This is where we override the marker to set it to false if realize we actually have the previous values in memory, meaning we want to update all columns (not just the `__artie_delete` column)

clients/bigquery/dialect/dialect.go
```diff
@@ -249,6 +249,11 @@ MERGE INTO %s %s USING %s AS %s ON %s`,
 		tableID.FullyQualifiedName(), constants.TargetAlias, subQuery, constants.StagingAlias, strings.Join(equalitySQLParts, " AND "),
 	)
 
+	cols, removed := columns.RemoveOnlySetDeleteColumnMarker(cols)
+	if !removed {
+		return []string{}, errors.New("artie only_set_delete flag doesn't exist")
+	}
```
danafallon, line 255: Bigquery isn't using this new marker yet, but still has to filter it out so it doesn't end up as a column in the destination table (same for mssql and redshift below)

clients/snowflake/dialect/dialect.go
```diff
@@ -225,20 +225,28 @@ MERGE INTO %s %s USING ( %s ) AS %s ON %s`,
...
-WHEN MATCHED %sTHEN UPDATE SET %s
+WHEN MATCHED %sAND IFNULL(%s, false) = false THEN UPDATE SET %s
+WHEN MATCHED %sAND IFNULL(%s, false) = true THEN UPDATE SET %s
 WHEN NOT MATCHED THEN INSERT (%s) VALUES (%s);`,
-			// Update + Soft Deletion
-			idempotentClause, sql.BuildColumnsUpdateFragment(cols, constants.StagingAlias, constants.TargetAlias, sd),
+			// Update + soft deletion when we have previous values (update all columns)
+			idempotentClause, sql.GetQuotedOnlySetDeleteColumnMarker(constants.StagingAlias, sd), sql.BuildColumnsUpdateFragment(cols, constants.StagingAlias, constants.TargetAlias, sd),
+			// Soft deletion when we don't have previous values (only update the __artie_delete column)
+			idempotentClause, sql.GetQuotedOnlySetDeleteColumnMarker(constants.StagingAlias, sd), sql.BuildColumnsUpdateFragment([]columns.Column{columns.NewColumn(constants.DeleteColumnMarker, typing.Boolean)}, constants.StagingAlias, constants.TargetAlias, sd),
```
danafallon, line 241: Here's the new merge logic - if `OnlySetDeleteColumnMarker` is set to true, then we only update the `__artie_delete` column instead of updating all columns.

models/event/event.go
```diff
@@ -69,6 +69,7 @@ func ToMemoryEvent(event cdc.Event, pkMap map[string]any, tc *kafkalib.TopicConf
 
 		// We don't need this either.
 		delete(evtData, constants.DeleteColumnMarker)
+		delete(evtData, constants.OnlySetDeleteColumnMarker)
```
danafallon, line 72: This removes the marker if we're in history mode, so that each dialect's `Append()` method doesn't have to filter it out like the `Merge()` methods do.

Tang8330: Can we add a test to this to ensure the column is being deleted? Can be a FF

clients/bigquery/dialect/dialect.go
```diff
@@ -249,6 +249,11 @@ MERGE INTO %s %s USING %s AS %s ON %s`,
...
+
 	if softDelete {
```
Tang8330, line 257: Leave a TODO to update the merge query?

clients/mssql/dialect/dialect.go
```diff
@@ -199,6 +199,11 @@ USING %s AS %s ON %s`,
 		strings.Join(sql.BuildColumnComparisons(primaryKeys, constants.TargetAlias, constants.StagingAlias, sql.Equal, md), " AND "),
 	)
 
+	cols, removed := columns.RemoveOnlySetDeleteColumnMarker(cols)
+	if !removed {
+		return []string{}, errors.New("artie only_set_delete flag doesn't exist")
+	}
+
```
Tang8330, line 206: Leave a TODO?

clients/redshift/dialect/dialect.go
```diff
@@ -257,6 +257,11 @@ func (rd RedshiftDialect) BuildMergeQueries(
 	// With AI, the sequence will increment (never decrement). And UUID is there to prevent universal hash collision
 	// However, there may be edge cases where folks end up restoring deleted rows (which will contain the same PK).
 
+	cols, removed := columns.RemoveOnlySetDeleteColumnMarker(cols)
```
Tang8330, line 260: Ditto - re: TODO

lib/cdc/mongo/debezium.go
Tang8330: Add a test?

lib/cdc/util/relational_event.go
Tang8330: Add test?

lib/config/constants/constants.go
```diff
@@ -15,6 +15,7 @@ const (
 	HistoryModeSuffix           = "__history"
 	ArtiePrefix                 = "__artie"
 	DeleteColumnMarker          = ArtiePrefix + "_delete"
+	OnlySetDeleteColumnMarker   = ArtiePrefix + "_only_set_delete"
```
Tang8330, line 18: Add a comment to this column?

Tang8330 (approved):

PR #801, opened 2024-07-18
Update redshift merge logic to not null out values when soft-deleting

danafallon:
In the same vein as #796 and #800.

Screenshot 2024-07-18 at 3 29 26 PM
^ row 1028 was deleted before these changes, so the other values were nulled out. Rows 1029-1031 were deleted with these changes, so the previous values were preserved. For rows 1030 & 1031, I included multiple operations in a single flush (c + u + d, u + d) to verify that the latest in-memory values were preserved.


Tang8330 (approved):
lgtm

PR #802, opened 2024-07-19
Update MS SQL merge logic to not null out values when soft-deleting

danafallon:
Had to use a different approach for this than for the other destinations, since MS SQL [doesn't allow](https://learn.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-ver16#when-matched-then-merge_matched) multiple `WHEN MATCHED THEN UPDATE` clauses. So instead, the `__artie_only_set_delete` flag is checked when determining the value to assign for each column (except the `__artie_delete` column). If `__artie_only_set_delete` is true, then it uses the target table's existing value; if false, it uses the new value from the staging table.

The update clause ends up looking like this:
```sql
WHEN MATCHED THEN UPDATE SET
  "id"=CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN stg."id" ELSE tgt."id" END,
  "first_name"=CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN stg."first_name" ELSE tgt."first_name" END,
  "last_name"=CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN stg."last_name" ELSE tgt."last_name" END,
  "email"=CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN stg."email" ELSE tgt."email" END,
  "__artie_updated_at"=CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN stg."__artie_updated_at" ELSE tgt."__artie_updated_at" END,
  "__artie_delete"=stg."__artie_delete"
```

### DIFF
--- a/clients/mssql/dialect/dialect.go
+++ b/clients/mssql/dialect/dialect.go
@@ -204,12 +204,24 @@ USING %s AS %s ON %s`,
 	}
 
 	if softDelete {
-		// TODO alter this merge query to update only the __artie_delete column if OnlySetDeleteColumnMarker is true
+		// Build conditional column values for the update clause; if __artie_only_set_delete is true, we
+		// need to preserve existing values for all columns other than __artie_delete
+		colsExceptDelete, err := columns.RemoveDeleteColumnMarker(cols)
+		if err != nil {
+			return nil, err
+		}
+
+		condition := fmt.Sprintf("COALESCE(%s, 0) = 0", sql.GetQuotedOnlySetDeleteColumnMarker(constants.StagingAlias, md))
+		conditionalCols := sql.BuildConditionalColumnsUpdateFragment(colsExceptDelete, constants.StagingAlias, constants.TargetAlias, md, condition)
+		deleteCol := sql.BuildColumnsUpdateFragment([]columns.Column{columns.NewColumn(constants.DeleteColumnMarker, typing.Boolean)}, constants.StagingAlias, constants.TargetAlias, md)
+
 		return []string{baseQuery + fmt.Sprintf(`
-WHEN MATCHED %sTHEN UPDATE SET %s
+WHEN MATCHED %sTHEN UPDATE SET %s,%s
 WHEN NOT MATCHED THEN INSERT (%s) VALUES (%s);`,
-			// WHEN MATCHED %sTHEN UPDATE SET %s
-			idempotentClause, sql.BuildColumnsUpdateFragment(cols, constants.StagingAlias, constants.TargetAlias, md),
+			// Updating or soft-deleting
+			// WHEN MATCHED %sTHEN UPDATE SET %s,%s
+			idempotentClause, conditionalCols, deleteCol,
+			// Inserting
 			// WHEN NOT MATCHED THEN INSERT (%s)
 			strings.Join(sql.QuoteColumns(cols, md), ","),
 			// VALUES (%s);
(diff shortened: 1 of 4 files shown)


clients/mssql/dialect/dialect_test.go
```diff
@@ -230,7 +230,28 @@ WHEN NOT MATCHED AND COALESCE(stg."__artie_delete", 1) = 0 THEN INSERT ("id","ba
...
+			_cols,
+			true,
+			false,
+		)
+		assert.NoError(t, err)
+		assert.Len(t, queries, 1)
+		assert.Equal(t, `
+MERGE INTO database.schema.table tgt
+USING {SUB_QUERY} AS stg ON tgt."id" = stg."id"
+WHEN MATCHED THEN UPDATE SET "id"= CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN stg."id" ELSE tgt."id" END,"bar"= CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN  CASE WHEN COALESCE(stg."bar", '') != '__debezium_unavailable_value' THEN stg."bar" ELSE tgt."bar" END ELSE tgt."bar" END,"updated_at"= CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN stg."updated_at" ELSE tgt."updated_at" END,"start"= CASE WHEN COALESCE(stg."__artie_only_set_delete", 0) = 0 THEN stg."start" ELSE tgt."start" END,"__artie_delete"=stg."__artie_delete"
```
danafallon, line 254: The assignment for the `bar` column ends up having nested case statements in it, so I wanted to add a test case for that. I ran it against my ms sql instance to verify that syntax works as expected.

clients/mssql/dialect/dialect.go
```diff
@@ -204,12 +204,24 @@ USING %s AS %s ON %s`,
...
+			return nil, err
+		}
+
+		condition := fmt.Sprintf("COALESCE(%s, 0) = 0", sql.GetQuotedOnlySetDeleteColumnMarker(constants.StagingAlias, md))
+		conditionalCols := sql.BuildConditionalColumnsUpdateFragment(colsExceptDelete, constants.StagingAlias, constants.TargetAlias, md, condition)
+		deleteCol := sql.BuildColumnsUpdateFragment([]columns.Column{columns.NewColumn(constants.DeleteColumnMarker, typing.Boolean)}, constants.StagingAlias, constants.TargetAlias, md)
+
 		return []string{baseQuery + fmt.Sprintf(`
-WHEN MATCHED %sTHEN UPDATE SET %s
+WHEN MATCHED %sTHEN UPDATE SET %s,%s
```
Tang8330, line 219: Can we return 2 updates and an insert similar to Redshift?

Tang8330: Btw, if you return multiple parts - Transfer will automatically detect that and wrap a tx: https://github.com/artie-labs/transfer/blob/879ae1212e440060537d1ce1c165d0bf46096233/lib/destination/dwh.go#L40

danafallon: Yeah that's an alternative. I think it will require a bigger refactor because we can't combine merge & update statements, right? So we'd have to also rewrite how we do inserts and deletes?

Out of curiosity, why do we use that approach (separate insert/update/delete statements instead of a single merge statement) for redshift? What was the limitation we ran into with redshift that a merge statement couldn't do?

Tang8330:
> Yeah that's an alternative. I think it will require a bigger refactor
because we can't combine merge & update statements, right? So we'd have to
also rewrite how we do inserts and deletes?

It shouldn't, you would only need to refactor this return statement.

> Out of curiosity, why do we use that approach (separate
insert/update/delete statements instead of a single merge statement) for
redshift? What was the limitation we ran into with redshift that a merge
statement couldn't do?

We did that with Redshift because Redshift's merge did not support multiple
case when statements

danafallon:
Closing in favor of https://github.com/artie-labs/transfer/pull/805

PR #810, opened 2024-07-22
[Table Config] Removing empty comment as a struct

Tang8330:
As part of simplifying the way we retrieve table config, this PR removes the need to specify what an "empty" comment looks like by having better parsing.


clients/shared/table_config.go
```diff
@@ -90,35 +80,41 @@ func (g GetTableCfgArgs) GetTableConfig() (*types.DwhTableConfig, error) {
 			return nil, err
 		}
 
-		row := make(map[string]string)
+		row := make(map[string]*string)
```
nathan-artie, line 83: Could we just store an empty string when we get a back a `nil` rather than use pointers?

Tang8330: Sure

nathan-artie (approved):

PR #812, opened 2024-07-24
[Debezium] Minor refactor around converters

Tang8330:
## Changes

1. Starting to stand up a `converters` pkg
2. Starting to pull logic out of the main `ParseValue` function, into smaller functions that are easier to unit test

### DIFF
--- a/lib/debezium/converters/time.go
+++ b/lib/debezium/converters/time.go
@@ -1,0 +1,35 @@
+package converters
+
+import (
+	"fmt"
+	"strings"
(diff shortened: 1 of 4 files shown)


lib/debezium/converters/time.go
```diff
@@ -0,0 +1,35 @@
+package converters
```
Tang8330, line 1: Just copied this from `types.go`

danafallon (approved):

PR #814, opened 2024-07-24
[Snowflake] Support authentication via private key

Tang8330:
We currently only support username and passwords.

We're having more customers ask for a key based authentication now. This PR adds the ability to authenticate via private keys.

### DIFF
--- a/lib/config/snowflake.go
+++ b/lib/config/snowflake.go
@@ -1,11 +1,9 @@
 package config
 
 import (
-	"crypto/rsa"
-	"crypto/x509"
-	"encoding/pem"
 	"fmt"
-	"os"
+
+	"github.com/artie-labs/transfer/lib/crypto"
 
 	"github.com/artie-labs/transfer/lib/ptr"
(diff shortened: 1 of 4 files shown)


lib/config/snowflake.go
```diff
@@ -0,0 +1,57 @@
...
+	if s.PathToPrivateKey != "" {
+		key, err := crypto.LoadRSAKey(s.PathToPrivateKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load private key: %w", err)
+		}
+
+		cfg.PrivateKey = key
+		cfg.Authenticator = gosnowflake.AuthTypeJwt
+	} else {
+		cfg.Password = s.Password
```
nathan-artie, line 47: Might be worth returning an error if both are empty strings.

nathan-artie (approved):

lib/config/snowflake.go
Tang8330: Snowflake's SDK will error out if it is empty

PR #816, opened 2024-07-24
[Debezium] Native Time with timezone parsing

Tang8330:
All the other time data types are already handled, it was just `TimeWithTimezone`

### DIFF
--- a/lib/debezium/converters/time_test.go
+++ b/lib/debezium/converters/time_test.go
@@ -65,4 +65,10 @@ func TestConvertTimeWithTimezone(t *testing.T) {
 
 		assert.Equal(t, expectedTs, ts)
 	}
+	{
+		// Non UTC
+		ts, err := ConvertTimeWithTimezone("23:02:06.745116")
+		assert.ErrorContains(t, err, `failed to parse "23:02:06.745116"`)
+		assert.Nil(t, ts)
+	}
 }
--- a/lib/debezium/converters/time_test.go
+++ b/lib/debezium/converters/time_test.go
@@ -46,7 +46,13 @@ func TestConvertDateTimeWithTimezone(t *testing.T) {
 
 func TestConvertTimeWithTimezone(t *testing.T) {
 	{
-		// What Debezium would produce
+		// Invalid
+		ts, err := ConvertTimeWithTimezone("23:02")
+		assert.Nil(t, ts)
+		assert.ErrorContains(t, err, `failed to parse "23:02": parsing time`)
+	}
+	{
+		// What Debezium + Reader would produce
 		ts, err := ConvertTimeWithTimezone("23:02:06.745116Z")
 		assert.NoError(t, err)
 		expectedTs := &ext.ExtendedTime{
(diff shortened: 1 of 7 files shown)


lib/typing/ext/time.go
```diff
@@ -82,11 +82,3 @@ func (e *ExtendedTime) String(overrideFormat string) string {
 
 	return e.Time.Format(e.NestedKind.Format)
 }
-
-func (e *ExtendedTime) StringUTC(overrideFormat string) string {
```
Tang8330, line 86: Unused

lib/typing/ext/parse.go
```diff
@@ -24,10 +25,25 @@ func ParseFromInterface(val any, additionalDateFormats []string) (*ExtendedTime,
 	return extendedTime, nil
 }
 
-// ParseTimeExactMatch is a wrapper around time.Parse() and will return an extra boolean to indicate if it was an exact match or not.
+// ParseTimeExactMatch - This function is the same as `ParseTimeExactMatchLegacy` with the only exception that it'll return if it was not an exact match
```
danafallon, line 28: "it'll return if it was not an exact match" - should this say "it'll return an error if it was not an exact match"?

lib/debezium/converters/time_test.go
```diff
@@ -43,3 +43,26 @@ func TestConvertDateTimeWithTimezone(t *testing.T) {
...
+func TestConvertTimeWithTimezone(t *testing.T) {
+	{
+		// Invalid
+		ts, err := ConvertTimeWithTimezone("23:02")
+		assert.Nil(t, ts)
+		assert.ErrorContains(t, err, `failed to parse "23:02": parsing time`)
+	}
+	{
+		// What Debezium + Reader would produce
+		ts, err := ConvertTimeWithTimezone("23:02:06.745116Z")
```
nathan-artie, line 56: Can we add a test for a non-UTC time zone?

lib/typing/ext/parse.go
```diff
@@ -24,10 +25,25 @@ func ParseFromInterface(val any, additionalDateFormats []string) (*ExtendedTime,
...
+func ParseTimeExactMatch(layout, potentialDateTimeString string) (time.Time, error) {
+	ts, err := time.Parse(layout, potentialDateTimeString)
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	if ts.Format(layout) != potentialDateTimeString {
```
nathan-artie, line 35: In what case would a formatted value not be the same as the value that was parsed? Is this to ensure that we are only matching time strings that have the same microsecond precision as the layout?

lib/debezium/converters/time_test.go
nathan-artie: Err sorry I meant can we add a test for a different time zone other than UTC, unless we only support parsing UTC?

Tang8330:
> In what case would a formatted value not be the same as the value that
was parsed? Is this to ensure that we are only matching time strings that
have the same microsecond precision as the layout?

For example:
1. potentialDateTimeString = 15:04:05.000000Z
2. layout = 15:04:05Z

It will parse correctly. However, if we went with this layout, we would
lose microsecond precision.

lib/debezium/converters/time_test.go
Tang8330: Added test case

lib/typing/ext/parse.go
Tang8330: Replied via email but the format got messed up.

Yes, it's to ensure that precision did not get truncated.

Tang8330: For example:
1. potentialDateTimeString = 15:04:05.000000Z
2. layout = 15:04:05Z

It will parse correctly. However, if we went with this layout, we would
lose microsecond precision.

nathan-artie: Gotcha, could we add a comment here to that effect?

nathan-artie (approved):

lib/typing/ext/parse.go
Tang8330: Yup

PR #820, opened 2024-07-28
[Debezium] Adding Geography + JSON converters

Tang8330:
Moving Geography and JSON to follow the converters pattern

### DIFF
--- a/lib/debezium/converters/geometry.go
+++ b/lib/debezium/converters/geometry.go
@@ -69,8 +69,14 @@ func (GeometryPoint) Convert(value any) (any, error) {
 	return string(bytes), nil
 }
 
-// ParseGeometry will take in an object with b64 encoded wkb and return a GeoJSON string.
-func ParseGeometry(value any) (string, error) {
+type Geography struct{}
+
+func (Geography) ToKindDetails() typing.KindDetails {
+	// We will return this in GeoJSON format.
+	return typing.Struct
+}
+
+func (Geography) Convert(value any) (any, error) {
 	valMap, isOk := value.(map[string]any)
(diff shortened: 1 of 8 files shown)


lib/debezium/converters/geometry.go
```diff
@@ -69,8 +69,14 @@ func (GeometryPoint) Convert(value any) (any, error) {
 	return string(bytes), nil
 }
 
-// ParseGeometry will take in an object with b64 encoded wkb and return a GeoJSON string.
-func ParseGeometry(value any) (string, error) {
+type Geography struct{}
```
danafallon, line 72: Is there a reason for calling this `Geography` instead of `Geometry` like it was before?

danafallon (approved):
One question about naming, otherwise looks good

lib/debezium/converters/geometry.go
Tang8330: Haha yeah I named it Geography because we already have a variable named Geometry

Tang8330: And this parser handles both Geography + Geometry

Tang8330: I could rename [Geometry](https://github.com/artie-labs/transfer/blob/875987a24aa5e4248ba4704c39583c56203ad6b6/lib/debezium/converters/geometry.go#L23) to be GeometryJSON.

Wdyt?

danafallon: Ah okay, yeah I think that seems more consistent/user-friendly to me (Geometry -> GeometryJSON and Geography -> Geometry)

PR #824, opened 2024-07-29
Tightening variable + tests

Tang8330:
## Change

1. Adding more tests and refactoring existing ones
2. `colKind` was actually just the `columns.Column` object before. We are now passing the actual `KindDetails` struct instead.

### DIFF
--- a/clients/redshift/cast.go
+++ b/clients/redshift/cast.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/artie-labs/transfer/lib/config/constants"
 	"github.com/artie-labs/transfer/lib/typing"
-	"github.com/artie-labs/transfer/lib/typing/columns"
 	"github.com/artie-labs/transfer/lib/typing/values"
 )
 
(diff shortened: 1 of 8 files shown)


clients/redshift/cast.go
```diff
@@ -34,11 +33,11 @@ func replaceExceededValues(colVal string, colKind columns.Column) string {
 	return colVal
 }
 
-// CastColValStaging - takes `colVal` any and `colKind` typing.Column and converts the value into a string value
+// castColValStaging - takes `colVal` any and `colKind` typing.Column and converts the value into a string value
 // This is necessary because CSV writers require values to in `string`.
-func (s *Store) CastColValStaging(colVal any, colKind columns.Column, additionalDateFmts []string) (string, error) {
+func castColValStaging(colVal any, colKind typing.KindDetails, additionalDateFmts []string) (string, error) {
```
Tang8330, line 38: Did not need to be on the store object, nor did it need to be public.

clients/snowflake/staging.go
```diff
@@ -43,7 +42,7 @@ func replaceExceededValues(colVal string, kindDetails typing.KindDetails) string
 
 // castColValStaging - takes `colVal` any and `colKind` typing.Column and converts the value into a string value
```
danafallon, line 43: Update this comment to say `typing.KindDetails` instead of `typing.Column`? (or just remove the types from the comment and let the function be the source of truth)

Tang8330: sg

clients/redshift/cast.go
```diff
@@ -5,25 +5,24 @@ import (
 
 	"github.com/artie-labs/transfer/lib/config/constants"
 	"github.com/artie-labs/transfer/lib/typing"
-	"github.com/artie-labs/transfer/lib/typing/columns"
 	"github.com/artie-labs/transfer/lib/typing/values"
 )
 
 const maxRedshiftLength int32 = 65535
 
 // replaceExceededValues - takes `colVal` any and `colKind` columns.Column and replaces the value with an empty string if it exceeds the max length.
```
danafallon, line 13: Oh this one also has the wrong type in it

Tang8330: Wow the comment for this one is verbose. Removing

danafallon (approved):

PR #825, opened 2024-07-29
MongoDB clean up

Tang8330:
As part of validating MongoDB streaming, I noticed we weren't handling `CodeWithScope` and also we were wrongfully assuming that BinData will be only 16 bytes. 

In MongoDB, you can create a BinData with any arbitrary value and assign whichever subtype you want.

### DIFF
--- a/lib/typing/mongo/bson.go
+++ b/lib/typing/mongo/bson.go
@@ -64,12 +64,17 @@ func bsonBinaryValueToMap(value primitive.Binary) (any, error) {
 	case
 		bson.TypeBinaryUUIDOld,
 		bson.TypeBinaryUUID:
-		parsedUUID, err := uuid.FromBytes(value.Data)
-		if err != nil {
-			return nil, err
+		if len(value.Data) == 16 {
+			// UUIDs must be 16 bytes
+			parsedUUID, err := uuid.FromBytes(value.Data)
+			if err != nil {
+				return nil, err
+			}
+
+			return parsedUUID.String(), nil
 		}
 
-		return parsedUUID.String(), nil
+		fallthrough
 	default:
 		return map[string]any{
 			"$binary": map[string]any{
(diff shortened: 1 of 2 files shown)


lib/typing/mongo/bson.go
```diff
@@ -103,12 +109,20 @@ func bsonValueToGoValue(value any) (any, error) {
...
+		return map[string]any{"$code": string(v.Code), "$scope": v.Scope}, nil
 	case
 		nil,
 		bool,
 		string,
 		int32, int64,
-		float32, float64:
+		float32:
+		return v, nil
+	case float64:
```
nathan-artie, line 121: Would we not want to check this for float32 too?

nathan-artie: Unless float32s can't be NaN/Inf?

lib/typing/mongo/bson.go
```diff
@@ -64,12 +65,17 @@ func bsonBinaryValueToMap(value primitive.Binary) (any, error) {
 	case
 		bson.TypeBinaryUUIDOld,
 		bson.TypeBinaryUUID:
-		parsedUUID, err := uuid.FromBytes(value.Data)
-		if err != nil {
-			return nil, err
+		if len(value.Data) == 16 {
```
nathan-artie, line 68: Are there UUIDs in MongoDB that are not 16 bytes or can these fields can be used to store non-UUID values?

Tang8330: The latter: can these fields can be used to store non-UUID values

lib/typing/mongo/bson.go
Tang8330: They theoretically can, but when you actually use the Infinity / -Infinity and NaN type, it creates float64

Tang8330: BSON parses doubles into float64s

nathan-artie (approved):

lib/typing/mongo/bson.go
Tang8330: To clarify, there should not be a way to get NaN / -/+ Inf that are float32 from BSON. But the language Go does technically allow it.

PR #826, opened 2024-07-30
[Typing] Remove the implicit JSON check

Tang8330:
We are already parsing JSON [here](https://github.com/artie-labs/transfer/blob/c01d76d8c3e9ab6506ffd19b65b7f5d5e38d1592/lib/debezium/types.go#L28). This PR removes the need for us to implicitly check if a string is a JSON object or not.

We will only label something as JSON if the schema provided, it or payload itself is a `map`.


lib/typing/values/string.go
```diff
@@ -57,22 +57,23 @@ func ToString(colVal any, colKind typing.KindDetails, additionalDateFmts []strin
...
-				colValBytes, err := json.Marshal(colVal)
-				if err != nil {
-					return "", err
-				}
+		if strings.Contains(fmt.Sprint(colVal), constants.ToastUnavailableValuePlaceholder) {
+			return fmt.Sprintf(`{"key":"%s"}`, constants.ToastUnavailableValuePlaceholder), nil
+		}
 
-				return string(colValBytes), nil
+		isArray := reflect.ValueOf(colVal).Kind() == reflect.Slice
```
Tang8330, line 64: Lifted this from `typing.String.Kind`

lib/typing/values/string.go
```diff
@@ -57,22 +57,23 @@ func ToString(colVal any, colKind typing.KindDetails, additionalDateFmts []strin
...
-					"key": constants.ToastUnavailableValuePlaceholder,
-				}
-			}
-
-			if reflect.TypeOf(colVal).Kind() != reflect.String {
-				colValBytes, err := json.Marshal(colVal)
-				if err != nil {
-					return "", err
-				}
+		if strings.Contains(fmt.Sprint(colVal), constants.ToastUnavailableValuePlaceholder) {
```
Tang8330, line 60: No need to create a map just to have it marshaled, we can do it ourselves. Also the if statement is a no-op given the case.

PR #827, opened 2024-07-30
[Debezium] Adding Decimal + Variable Decimal Converters

Tang8330:
This PR moves the existing decimal + variable decimal parsing functions into our converters framework


lib/debezium/converters/decimal_test.go
```diff
@@ -0,0 +1,224 @@
+package converters
+
+import (
+	"testing"
+
+	"github.com/artie-labs/transfer/lib/typing/decimal"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestToBytes(t *testing.T) {
```
Tang8330, line 10: This was just m oved

lib/debezium/schema.go
```diff
@@ -80,77 +82,83 @@ func (f Field) GetScaleAndPrecision() (int32, *int32, error) {
...
-	case Timestamp, MicroTimestamp, NanoTimestamp, DateTimeKafkaConnect:
-		return typing.NewKindDetailsFromTemplate(typing.ETime, ext.DateTimeKindType)
-	case MicroTime, NanoTime:
-		return typing.NewKindDetailsFromTemplate(typing.ETime, ext.TimeKindType)
+		return converters.Time{}, nil
 	case KafkaDecimalType:
 		scale, precisionPtr, err := f.GetScaleAndPrecision()
 		if err != nil {
-			return typing.Invalid
+			return nil, fmt.Errorf("failed to get scale and precision: %w", err)
```
Tang8330, line 104: Instead of returning `typing.Invalid`, let's explicitly fail.

lib/debezium/types.go
```diff
@@ -58,28 +56,6 @@ const (
 	KafkaDecimalPrecisionKey = "connect.decimal.precision"
 )
 
-// toBytes attempts to convert a value (type []byte, or string) to a slice of bytes.
-// - If value is already a slice of bytes it will be directly returned.
-// - If value is a string we will attempt to base64 decode it.
-func toBytes(value any) ([]byte, error) {
```
Tang8330, line 64: Moved

lib/debezium/converters/decimal.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/artie-labs/transfer/lib/maputil"
+	"github.com/artie-labs/transfer/lib/typing"
+	"github.com/artie-labs/transfer/lib/typing/decimal"
+)
+
+// toBytes attempts to convert a value (type []byte, or string) to a slice of bytes.
+// - If value is already a slice of bytes it will be directly returned.
+// - If value is a string we will attempt to base64 decode it.
+func toBytes(value any) ([]byte, error) {
```
Tang8330, line 17: Moved, no changes

lib/debezium/types.go
nathan-artie: I think https://github.com/artie-labs/transfer/pull/827/files#diff-6f8a144d4abeb92f8240e55383e15b8323e83bb3b9f69088274ed61f4f344e96R117 can probably go away too.

nathan-artie: The reason `toBytes` is here is because I was going to use it in this [switch statement](https://github.com/artie-labs/transfer/pull/827/files#diff-6f8a144d4abeb92f8240e55383e15b8323e83bb3b9f69088274ed61f4f344e96R85) so that if `Field.Type` is  `Bytes` we do the base64 decoding once at the field level and then just pass bytes to the individual value converter, similar to how we do float to integer conversion for the integer types. 

The goal was to be able to get rid of [this comment](https://github.com/artie-labs/transfer/pull/827/files#diff-6f8a144d4abeb92f8240e55383e15b8323e83bb3b9f69088274ed61f4f344e96R119) once we'd verified that bytes worked everywhere downstream.

Tang8330: Posting here for posterity as I'm not going to get to it right away.

The idea here is that we should handle the `toBytes` in the pre-processing layer, when we detect the incoming payload is `bytes`.

We can do that as a first pass, then the Debezium converters can expect bytes

PR #829, opened 2024-07-30
[MongoDB] Remove calling `typing.ParseValue`

Tang8330:
As part of our MongoDB streaming validation, I noticed that we were calling `typing.ParseValue` with the purpose of checking if the field is a struct or not. 

Instead, we can just call rely on reflect and remove the need to call `ParseValue`, which also simplifies the signature.

### DIFF
--- a/lib/cdc/mongo/debezium_test.go
+++ b/lib/cdc/mongo/debezium_test.go
@@ -10,10 +10,8 @@ import (
 	"github.com/artie-labs/transfer/lib/debezium"
 	"github.com/artie-labs/transfer/lib/typing"
 
-	"github.com/stretchr/testify/assert"
-	"go.mongodb.org/mongo-driver/bson"
-
 	"github.com/artie-labs/transfer/lib/kafkalib"
+	"github.com/stretchr/testify/assert"
 )
 
 func (m *MongoTestSuite) TestGetPrimaryKey() {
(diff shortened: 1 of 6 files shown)


lib/cdc/mongo/debezium_test.go
```diff
@@ -84,16 +80,6 @@ func (m *MongoTestSuite) TestSource_GetExecutionTime() {
 		18, 6, 35, 21, 0, time.UTC), schemaEvtPayload.GetExecutionTime())
 }
 
-func (m *MongoTestSuite) TestBsonTypes() {
```
nathan-artie, line 87: Intentional?

Tang8330: Yup, it was testing the bson library

processes/consumer/process.go
```diff
@@ -51,8 +51,7 @@ func (p processArgs) process(ctx context.Context, cfg config.Config, inMemDB *mo
 		return "", fmt.Errorf("cannot unmarshall key %s: %w", string(p.Msg.Key()), err)
 	}
 
-	typingSettings := cfg.SharedTransferConfig.TypingSettings
-	_event, err := topicConfig.GetEventFromBytes(typingSettings, p.Msg.Value())
+	_event, err := topicConfig.GetEventFromBytes(p.Msg.Value())
```
nathan-artie, line 54: Could change `process` to accept `cfg.Mode` instead of `cfg` now.

nathan-artie (approved):

processes/consumer/process.go
Tang8330: Ah, we still need it for event save

```go
	shouldFlush, flushReason, err := evt.Save(cfg, inMemDB, topicConfig.tc, p.Msg)
	if err != nil {
		tags["what"] = "save_fail"
		return "", fmt.Errorf("event failed to save: %w", err)
	}
```

nathan-artie: Ah missed that!

PR #834, opened 2024-08-13
[MSSQL] Don't rely on `ParseFromInterface`

Tang8330:
The function `ParseFromInterface` will call `ParseExtendedDateTime` which we are trying to deprecate. 

Starting to refactor the callers, starting with MSSQL.


nathan-artie (approved):

Tang8330:
Just tried testing this, if we receive the date time string, we need to parse it and then return it back in a format that MSSQL accepts

Tang8330:
Or give it back as a `time.Time` object and have them parse it.

Tang8330:
Going to try another approach

PR #836, opened 2024-08-13
[DateTime] Return `ext.ExtendedTime` as opposed to `string`

Tang8330:
Right now, we're returning `updated_at` back as a string. This then leads us to do additional parsing downstream

### DIFF
--- a/lib/cdc/mongo/debezium.go
+++ b/lib/cdc/mongo/debezium.go
@@ -184,7 +184,7 @@ func (s *SchemaEventPayload) GetData(pkMap map[string]any, tc *kafkalib.TopicCon
 	}
 
 	if tc.IncludeArtieUpdatedAt {
-		retMap[constants.UpdateColumnMarker] = ext.NewUTCTime(ext.ISO8601)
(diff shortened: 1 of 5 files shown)


danafallon:
> This then leads us to do additional parsing downstream

Where does that parsing happen? Is there any code we can remove there?

danafallon (approved):

Tang8330:
> Where does that parsing happen? Is there any code we can remove there?

When we write the staging tables out, we'll call [ParseFromInterface](https://github.com/artie-labs/transfer/blob/22d1e1281e4b99e8005352b68a8280c42023233d/lib/typing/ext/parse.go#L9).

We can't remove the code because we need it to parse `time.Time` from potential string values we receive from the source.

Tang8330:
An example of this would be DynamoDB where it doesn't have a data type for `time.Time`, but our customers can create the column in Snowflake as `TIMESTAMP` and Transfer will automatically pick that up and try to parse all the string values into a `TIMESTAMP`.

PR #842, opened 2024-08-14
More guardrails around `ShouldDeleteFromName`

Tang8330:
Similar logic to https://github.com/artie-labs/transfer/blob/060ed44aa26a31af73c38a0b55e29a74626d4eef/lib/destination/ddl/ddl.go#L22

### DIFF
--- a/lib/destination/ddl/expiry.go
+++ b/lib/destination/ddl/expiry.go
@@ -10,7 +10,7 @@ import (
 )
 
 func ShouldDeleteFromName(name string) bool {
-	if !strings.Contains(name, constants.ArtiePrefix) {
+	if !strings.Contains(strings.ToLower(name), constants.ArtiePrefix) {
 		return false
 	}
 
--- a/lib/destination/ddl/expiry.go
+++ b/lib/destination/ddl/expiry.go
@@ -5,9 +5,15 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/artie-labs/transfer/lib/config/constants"
 )
 
 func ShouldDeleteFromName(name string) bool {
+	if !strings.Contains(name, constants.ArtiePrefix) {
+		return false
+	}
+
 	nameParts := strings.Split(name, "_")
 	if len(nameParts) < 2 {
 		return false
--- a/lib/destination/ddl/expiry.go
+++ b/lib/destination/ddl/expiry.go
@@ -20,13 +20,12 @@ func ShouldDeleteFromName(name string) bool {
 		return false
 	}
 
-	tsString := suffixParts[len(suffixParts)-1]
-	unix, err := strconv.Atoi(tsString)
+	unix, err := strconv.Atoi(suffixParts[2])
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lib/destination/ddl/expiry.go
```diff
@@ -5,11 +5,22 @@ import (
...
 
 func ShouldDeleteFromName(name string) bool {
+	// We expect the table name to be in the format of `tableName__artie_suffix_unix`
+	if !strings.Contains(strings.ToLower(name), constants.ArtiePrefix) {
+		return false
+	}
+
 	nameParts := strings.Split(name, "_")
-	if len(nameParts) < 2 {
+	if len(nameParts) < 6 {
```
danafallon, line 19: Why 6?

Tang8330: That's just how many are in a temp table

Tang8330: At the minimum

Tang8330: I could put it back to 2 if it's confusing?

danafallon: Oh I see, because right after the table name there are 3 underscores in a row, so 2 of the 6 parts are empty. Maybe it would be more clear to just split the part that comes after the `constants.ArtiePrefix` and check for 2 parts there? Because otherwise if the table name itself has any underscores in it, we could get a false positive here.

lib/destination/ddl/expiry_test.go
```diff
@@ -2,29 +2,43 @@ package ddl
...
-
-	for _, tblToNotDelete := range tblsToNotDelete {
-		assert.False(t, ShouldDeleteFromName(tblToNotDelete), tblToNotDelete)
-	}
+	{
+		// Tables to not drop
+		tablesToNotDrop := []string{
+			"foo",
+			"transactions",
+			fmt.Sprintf("expired_tbl__artie_suffix_%d", time.Now().Add(constants.TemporaryTableTTL).Unix()),
```
danafallon, line 20: Call this `future_tbl__...` for clarity?

lib/destination/ddl/expiry.go
```diff
@@ -5,21 +5,31 @@ import (
...
+	}
+
+	suffixParts := strings.Split(parts[1], "_")
+	if len(suffixParts) != 3 {
 		return false
 	}
 
-	unixString := nameParts[len(nameParts)-1]
-	unix, err := strconv.Atoi(unixString)
+	tsString := suffixParts[len(suffixParts)-1]
```
danafallon, line 23: Could just hardcode this to `[2]` if we're enforcing that `len(suffixParts)` has to be exactly 3, but I think it's also fine as is.

danafallon (approved):

lib/destination/ddl/expiry_test.go
Tang8330: Yup!

lib/destination/ddl/expiry.go
Tang8330: True

PR #843, opened 2024-08-16
Simplifying Flush

Tang8330:
## Scope

Simplifying our flush function to do the following:

1. Get rid of the concept of "retryable errors", this approach doesn't scale (There are far too many errors, and we'd be spending most of our time trying to chase down edge cases)

2. Split Flush into two separate function and implement the Retryable framework

The intended outcome of this PR is to have the ability to retry a few times (with some jitter) after we first fail to merge or append. The previous approach would wait until the next flush cycle to retry, rather than retry right away.

By doing this, this should help alleviate intermittent errors and make our monitor less flakey.

### DIFF
--- a/lib/retry/retry.go
+++ b/lib/retry/retry.go
@@ -66,6 +66,7 @@ func sleepIfNecessary(cfg RetryConfig, attempt int, err error) {
 
 // WithRetries runs function `f` and returns the error if one occurs.
 func WithRetries(cfg RetryConfig, f func(attempt int, err error) error) error {
+	var prevErr error
 	var err error
 	for attempt := 0; attempt < cfg.MaxAttempts(); attempt++ {
 		sleepIfNecessary(cfg, attempt, err)
@@ -73,13 +74,15 @@ func WithRetries(cfg RetryConfig, f func(attempt int, err error) error) error {
 		if err == nil {
 			if attempt > 0 {
 				// Only log if there was more than one attempt, so it's less noisy.
-				slog.Info("Retry was successful!", slog.Int("attempts", attempt))
+				slog.Info("Retry was successful!", slog.Int("attempts", attempt), slog.Any("prevErr", prevErr))
 			}
 
 			return nil
 		} else if !cfg.IsRetryableErr(err) {
 			break
 		}
+
+		prevErr = err
 	}
 
 	return err
--- a/lib/retry/retry.go
+++ b/lib/retry/retry.go
@@ -71,11 +71,17 @@ func WithRetries(cfg RetryConfig, f func(attempt int, err error) error) error {
 		sleepIfNecessary(cfg, attempt, err)
 		err = f(attempt, err)
 		if err == nil {
+			if attempt > 0 {
+				// Only log if there was more than one attempt, so it's less noisy.
+				slog.Info("Retry was successful!", slog.Int("attempts", attempt))
+			}
+
 			return nil
 		} else if !cfg.IsRetryableErr(err) {
 			break
 		}
 	}
+
 	return err
 }
 
--- a/processes/consumer/flush.go
+++ b/processes/consumer/flush.go
@@ -67,7 +67,7 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
 				action = "append"
 			}
 
-			retryCfg, err := retry.NewJitterRetryConfig(250, 15_000, 5, retry.AlwaysRetry)
+			retryCfg, err := retry.NewJitterRetryConfig(500, 30_000, 10, retry.AlwaysRetry)
 			if err != nil {
 				slog.Error("Failed to create retry config", slog.Any("err", err))
 				return
--- a/processes/consumer/flush.go
+++ b/processes/consumer/flush.go
@@ -129,7 +129,8 @@ func flush(ctx context.Context, dest destination.Baseline, metricsClient base.Cl
 	}
 
 	if err = commitOffset(ctx, _tableData.TopicConfig().Topic, _tableData.PartitionsToLastMessage); err != nil {
-		return fmt.Errorf("failed to commit offset: %w", err)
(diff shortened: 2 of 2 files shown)


lib/retry/retry.go
```diff
@@ -71,11 +71,17 @@ func WithRetries(cfg RetryConfig, f func(attempt int, err error) error) error {
 		sleepIfNecessary(cfg, attempt, err)
 		err = f(attempt, err)
 		if err == nil {
+			if attempt > 0 {
+				// Only log if there was more than one attempt, so it's less noisy.
+				slog.Info("Retry was successful!", slog.Int("attempts", attempt))
```
nathan-artie, line 76: Might be worth logging the last error too.

processes/consumer/flush.go
```diff
@@ -85,35 +89,46 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
...
-				slog.With(logFields...).Error(fmt.Sprintf("Failed to execute %s, not going to flush memory, will sleep for 3 seconds before continuing...", action), slog.Any("err", err))
-				time.Sleep(3 * time.Second)
+				slog.Error(fmt.Sprintf("Failed to %s", action), slog.Any("err", err), slog.String("tableName", _tableName))
 			} else {
-				slog.Info(fmt.Sprintf("%s success, clearing memory...", stringutil.CapitalizeFirstLetter(action)), logFields...)
-				commitErr := commitOffset(ctx, _tableData.TopicConfig().Topic, _tableData.PartitionsToLastMessage)
-				if commitErr == nil {
-					inMemDB.ClearTableConfig(_tableName)
-				} else {
-					tags["what"] = "commit_fail"
```
nathan-artie, line 109: If we are permanently unable to commit the offsets to Kafka due to either a misconfiguration or a bug in the code, are we going to get notified about it since we're not logging `commit_fail` anymore?

processes/consumer/flush.go
```diff
@@ -85,35 +89,46 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
...
-			var err error
-			action := "merge"
-			// Merge or Append depending on the mode.
-			if _tableData.Mode() == config.History {
-				err = dest.Append(_tableData.TableData, false)
-				action = "append"
-			} else {
-				err = dest.Merge(_tableData.TableData)
-			}
+			err = retry.WithRetries(retryCfg, func(_ int, _ error) error {
```
nathan-artie, line 92: If you capture `attempts` and store it in a variable you could add it to `tags` so we can track attempts in Datadog.

processes/consumer/flush.go
nathan-artie: Is there a risk that if we fail to commit offsets for an extended period of time when transfer eventually restarts it will have to pick up from an offset that is significantly in the past?

processes/consumer/flush.go
Tang8330: I thought about that too, but we can get the attempts into Datadog through the logging statement above. 

This would then not incur additional unique metric count. We also don't need granularity of the attempts at the Kafka topic config level as it's going to be used on aggregate to inform us on the retry configs.

lib/retry/retry.go
Tang8330: Sure

processes/consumer/flush.go
Tang8330: > If we are permanently unable to commit the Kafka offsets due to either a misconfiguration or a bug in the code, are we going to get notified about it if we're not reporting commit_fail metrics anymore?

We still can since warnings go into Datadog too. We could set up an alert there if we wanted.

> Is there a risk that if we fail to commit offsets for an extended period of time when transfer eventually restarts it will have to pick up from an offset that is significantly in the past?

Hm, it is possible, although I haven't seen it before. (We average ~1-3 occasional commit failures per month).

processes/consumer/flush.go
```diff
@@ -56,64 +57,76 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
...
+	} else {
+		err = dest.Merge(_tableData.TableData)
+	}
+
+	if err != nil {
+		return "merge_fail", fmt.Errorf("failed to flush: %w", err)
+	}
+
+	if err = commitOffset(ctx, _tableData.TopicConfig().Topic, _tableData.PartitionsToLastMessage); err != nil {
+		return "commit_fail", fmt.Errorf("failed to commit kafka offset: %w", err)
```
Tang8330, line 128: If Kafka commit fails, this will now mark the flush as errored `what = "commit_fail"` which would prevent `inMemDB.ClearTableConfig(_tableName)`

processes/consumer/flush.go
nathan-artie: I like it, preserving the existing behavior seems like the safer approach to me.

processes/consumer/flush.go
```diff
@@ -56,64 +57,76 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
...
-				slog.Info(fmt.Sprintf("%s success, clearing memory...", stringutil.CapitalizeFirstLetter(action)), logFields...)
-				commitErr := commitOffset(ctx, _tableData.TopicConfig().Topic, _tableData.PartitionsToLastMessage)
-				if commitErr == nil {
-					inMemDB.ClearTableConfig(_tableName)
-				} else {
-					tags["what"] = "commit_fail"
-					slog.Warn("Commit error...", slog.Any("err", commitErr))
-				}
+				inMemDB.ClearTableConfig(_tableName)
+				slog.Info(fmt.Sprintf("%s success, clearing memory...", stringutil.CapitalizeFirstLetter(action)), slog.String("tableName", _tableName))
```
nathan-artie, line 99: Might want to swap the order of these two lines, or change the message, as "clearing memory..." sort of implies that the memory hasn't been cleared yet.

nathan-artie (approved):

PR #850, opened 2024-08-24
[Debezium] Better handle on setting default values

Tang8330:
The purpose of this PR is to start filtering and logging certain default values from being applied.

We already know that values like `GEN_RANDOM_UUID()` or `CURRENT_TIMESTAMP()` cannot be replicated over. This PR filters that and starts logging things that we have not explicitly added as a case statement.


lib/debezium/types.go
```diff
@@ -99,6 +100,27 @@ func toInt64(value any) (int64, error) {
...
 
+// ShouldSetDefaultValue will filter out computed fields that cannot be properly set with a default value
+func (f Field) ShouldSetDefaultValue(defaultValue any) bool {
+	switch castedDefaultValue := defaultValue.(type) {
+	case nil:
+		return false
+	case *ext.ExtendedTime:
+		return !castedDefaultValue.Time.IsZero()
+	case string:
+		if f.DebeziumType == UUID && castedDefaultValue == "00000000-0000-0000-0000-000000000000" {
```
danafallon, line 111: Could we use `uuid.Nil` here instead of hardcoding, to make sure we have the right number of 0s?

danafallon (approved):

lib/debezium/types.go
Tang8330: Sure, I was debating doing `uuid.Nil.String()` or hardcoding the literal value. I'll add `uuid.Nil` back so it's more clear

PR #853, opened 2024-08-26
Enabling go vet

Tang8330:
Fixing errors and adding `make vet`


nathan-artie (approved):
Do we also want to run this when we run tests?

Tang8330:
> Do we also want to run this when we run tests?

SGTM

PR #866, opened 2024-09-01
Refactor schema tests

Tang8330:
Refactoring tests

### DIFF
--- a/lib/debezium/schema_test.go
+++ b/lib/debezium/schema_test.go
@@ -105,7 +105,7 @@ func TestField_ToKindDetails(t *testing.T) {
 			// Variable numeric decimal
 			assert.Equal(
 				t, typing.NewDecimalDetailsFromTemplate(typing.EDecimal, decimal.NewDetails(decimal.PrecisionNotSpecified, decimal.DefaultScale)),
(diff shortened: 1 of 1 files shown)


lib/debezium/schema_test.go
```diff
@@ -82,203 +79,85 @@ func TestField_GetScaleAndPrecision(t *testing.T) {
...
+	}
+	{
+		// Time
+		for _, dbzType := range []SupportedDebeziumType{Time, TimeKafkaConnect, MicroTime, NanoTime, TimeWithTimezone} {
+			assert.Equal(t, typing.NewKindDetailsFromTemplate(typing.ETime, ext.TimeKindType), Field{DebeziumType: dbzType}.ToKindDetails())
+		}
+	}
+	{
+		// Datetime
+		assert.Equal(t, typing.NewKindDetailsFromTemplate(typing.ETime, ext.DateTimeKindType), Field{DebeziumType: DateTimeWithTimezone}.ToKindDetails())
```
danafallon, line 161: Could this one go in the array on line 143?

Tang8330: Yeah we can put it in there. When we delineate support for TIMESTAMP and DATETIME, we can then pull it out

PR #868, opened 2024-09-03
Add warning for unhandled field / debezium type.

Tang8330:
We should be able to natively handle every single Debezium or native type. This log will help us identify any unhandled types (if any).

### DIFF
--- a/lib/debezium/schema.go
+++ b/lib/debezium/schema.go
@@ -1,6 +1,8 @@
 package debezium
 
 import (
+	"log/slog"
+
 	"github.com/artie-labs/transfer/lib/debezium/converters"
 	"github.com/artie-labs/transfer/lib/maputil"
 	"github.com/artie-labs/transfer/lib/ptr"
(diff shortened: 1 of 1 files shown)


lib/debezium/schema.go
```diff
@@ -155,6 +157,10 @@ func (f Field) ToKindDetails() typing.KindDetails {
 	case Array:
 		return typing.Array
 	default:
+		if f.Type != "" || f.DebeziumType != "" {
```
nathan-artie, line 160: Are `f.Type` or `f.DebeziumType` ever empty strings?

Tang8330: We did have a case of it being empty here: https://github.com/artie-labs/reader/pull/473

nathan-artie (approved):

lib/debezium/schema.go
Tang8330: Just thinking more about this, my eventual goal is to have this return an error instead of `typing.Invalid`.

Doing conditional checks on empty string will still require me to check who is passing empty strings. I just removed it so I can track down any potential codepaths that are emitting empty or unhandled types.

nathan-artie: Would it be useful to also log if `f.DebeziumType` is not empty after the `switch f.DebeziumType` statement has completed? In other words, if `f.DebeziumType` is present do we always expect that the case statement should handle it? Should it be an error if `f.DebeziumType` = `"junk"` and `f.Type` = "string"?

Tang8330: I'm down

Tang8330: Let's log it.

> Should it be an error if f.DebeziumType = "junk" and f.Type = "string"?

I think so

nathan-artie (approved):

PR #881, opened 2024-09-04
[Debezium] Moving Variable Decimal

Tang8330:
Moving the variable decimal converter. Once this is done, we only have one more and we're done!


lib/debezium/types_test.go
```diff
@@ -500,62 +500,3 @@ func TestField_DecodeDecimal(t *testing.T) {
 		assert.Equal(t, testCase.expectedScale, dec.Details().Scale(), testCase.name)
 	}
 }
-
-func TestField_DecodeDebeziumVariableDecimal(t *testing.T) {
```
Tang8330, line 504: Moved

nathan-artie (approved):

PR #883, opened 2024-09-05
[Typing] Stop implicit references of datetime in ParseValue

Tang8330:
All date, time, datetime and timestamp values are natively handled in the BSON and Debezium converters.

There is no need to do this anymore.

### DIFF
--- a/lib/typing/parse.go
+++ b/lib/typing/parse.go
@@ -14,10 +14,6 @@ func ParseValue(_ Settings, key string, optionalSchema map[string]KindDetails, v
 		return kindDetail
 	}
 
-	return parseValue(val)
-}
-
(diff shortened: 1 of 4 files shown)


nathan-artie (approved):

lib/typing/parse.go
```diff
@@ -4,21 +4,20 @@ import (
...
 	if kindDetail, isOk := optionalSchema[key]; isOk {
 		return kindDetail
 	}
 
-	return parseValue(settings, val)
+	return parseValue(val)
 }
 
-func parseValue(settings Settings, val any) KindDetails {
+func parseValue(val any) KindDetails {
```
nathan-artie, line 20: Could make this one function.

PR #887, opened 2024-09-05
[ON HOLD] Adding a function `SupportDefaultValue`

Tang8330:
Snowflake doesn't support it well, so let's remove support.

### DIFF
--- a/clients/shared/default_value.go
+++ b/clients/shared/default_value.go
@@ -5,7 +5,6 @@ import (
 	"log/slog"
 
 	bigQueryDialect "github.com/artie-labs/transfer/clients/bigquery/dialect"
-	mssqlDialect "github.com/artie-labs/transfer/clients/mssql/dialect"
 	"github.com/artie-labs/transfer/lib/destination"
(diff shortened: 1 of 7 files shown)


clients/shared/default_value.go
```diff
@@ -58,8 +57,7 @@ func BackfillColumn(dwh destination.DataWarehouse, column columns.Column, tableI
 		return nil
 	}
 
-	if _, ok := dwh.Dialect().(mssqlDialect.MSSQLDialect); ok {
-		// TODO: Support MSSQL column backfill
+	if !dwh.SupportDefaultValue() {
```
danafallon, line 60: In this case, should we continue backfilling with default values the old way? And only alter the column default (in https://github.com/artie-labs/transfer/pull/885) if this is true?

Tang8330: I think we should get rid of our backfill completely because it's error prone

danafallon (approved):

Tang8330:
Removing this

PR #888, opened 2024-09-05
[Debezium] Standing up Decimal converters - Part one

Tang8330:
First pass at standing up the Decimal converter

### DIFF
--- a/lib/debezium/schema.go
+++ b/lib/debezium/schema.go
@@ -129,9 +129,13 @@ func (f Field) ToValueConverter() (converters.ValueConverter, error) {
 		return converters.MicroTimestamp{}, nil
 	case NanoTimestamp:
 		return converters.NanoTimestamp{}, nil
-	}
+	default:
+		if f.DebeziumType != "" {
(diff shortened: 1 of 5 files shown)


lib/debezium/schema.go
```diff
@@ -81,70 +81,69 @@ func (f Field) GetScaleAndPrecision() (int32, *int32, error) {
...
-		if precisionPtr != nil {
-			precision = *precisionPtr
-		}
-
-		return typing.NewDecimalDetailsFromTemplate(typing.EDecimal, decimal.NewDetails(precision, scale))
+	if converter != nil {
+		return converter.ToKindDetails()
 	}
 
 	if f.DebeziumType != "" {
```
nathan-artie, line 149: Could move this to the `default` case of the switch statement in `ToValueConverter()`.

nathan-artie (approved):

lib/debezium/schema.go
Tang8330: sgtm

PR #890, opened 2024-09-06
[Debezium] Implementing bits + other string types.

Tang8330:
Implementing the following:

* Bit (parsing as `[]byte` and returning b64 encoded string)
* Interval and XML string passthroughs

### DIFF
--- a/lib/debezium/converters/basic.go
+++ b/lib/debezium/converters/basic.go
@@ -42,14 +42,14 @@ func (Int64Passthrough) Convert(value any) (any, error) {
 	return value, nil
 }
 
-type Bit struct{}
+type Base64 struct{}
 
-func (Bit) ToKindDetails() typing.KindDetails {
+func (Base64) ToKindDetails() typing.KindDetails {
 	// We're returning this back as a base64 encoded string.
 	return typing.String
 }
 
-func (Bit) Convert(value any) (any, error) {
+func (Base64) Convert(value any) (any, error) {
 	castedValue, err := typing.AssertType[[]byte](value)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


lib/debezium/types.go
```diff
@@ -28,11 +28,16 @@ const (
...
+	JSON     SupportedDebeziumType = "io.debezium.data.Json"
+	Enum     SupportedDebeziumType = "io.debezium.data.Enum"
+	EnumSet  SupportedDebeziumType = "io.debezium.data.EnumSet"
+	UUID     SupportedDebeziumType = "io.debezium.data.Uuid"
+	LTree    SupportedDebeziumType = "io.debezium.data.Ltree"
+	Interval SupportedDebeziumType = "io.debezium.time.Interval"
+	XML      SupportedDebeziumType = "io.debezium.data.Xml"
+
+	// Bytes
+	Bit SupportedDebeziumType = "io.debezium.data.Bits"
```
nathan-artie, line 40: Nit can we call this `Bits`? `Bit` makes me think this is just a single bit.

lib/debezium/converters/basic.go
```diff
@@ -40,3 +41,19 @@ func (Int64Passthrough) Convert(value any) (any, error) {
 
 	return value, nil
 }
+
+type Bit struct{}
```
nathan-artie, line 45: Could call this `Base64` since it's a converter that produces base64 encoded values.

nathan-artie (approved):

lib/debezium/types.go
Tang8330: Yeah sure

lib/debezium/converters/basic.go
Tang8330: sgtm

PR #896, opened 2024-09-06
[Typing] Simplify Ext Parse

Tang8330:
Will merge this PR based on the findings of: https://github.com/artie-labs/transfer/pull/895

### DIFF
--- a/lib/typing/ext/parse.go
+++ b/lib/typing/ext/parse.go
@@ -2,7 +2,6 @@ package ext
 
 import (
 	"fmt"
-	"log/slog"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


lib/typing/ext/parse.go
```diff
@@ -62,48 +48,27 @@ func ParseTimeExactMatchLegacy(layout, potentialDateTimeString string) (time.Tim
...
 			return NewExtendedTime(ts, DateKindType, supportedDateFormat), nil
 		}
 	}
 
-	// TODO: Remove this if we don't see any Sentry.
 	// Now check TIME formats
 	for _, supportedTimeFormat := range SupportedTimeFormatsLegacy {
-		ts, exactMatch, err := ParseTimeExactMatchLegacy(supportedTimeFormat, dtString)
-		if err == nil && exactMatch {
-			slog.Error("Unexpected call to SupportedTimeFormatsLegacy",
```
nathan-artie, line 93: Should we remove this entire loop? Looks like we don't actually expect to match any time formats, but with this change we'll still support matching time formats.

Tang8330: We should keep it because we'll need it for the inference work later (When we encounter a `TIME` column later and we have a string value, we should try to parse the string value as `TIME`)

nathan-artie (approved):

PR #902, opened 2024-09-10
[Typing] ParseFromInterface to take ExtendedTimeKindType

Tang8330:
In this PR, we are refactoring `ParseFromInterface` to take `ExtendedTimeKindType` so we know what we need to be parsing against.

This way, for `DateTime` types, we don't need to try and parse layouts that are targeted for `DATE` and `TIME`.

### DIFF
--- a/lib/typing/ext/parse.go
+++ b/lib/typing/ext/parse.go
@@ -20,14 +20,14 @@ func ParseTimeExactMatch(layout, value string) (time.Time, error) {
 	return ts, nil
 }
 
-func ParseFromInterfaceNew(val any, kindType ExtendedTimeKindType) (*ExtendedTime, error) {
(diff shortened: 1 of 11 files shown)


lib/typing/ext/parse.go
```diff
@@ -23,42 +38,63 @@ func ParseFromInterface(val any) (*ExtendedTime, error) {
...
 }
 
-// ParseTimeExactMatch will return an error if it was not an exact match.
-// We need this function because things may parse correctly but actually truncate precision
-func ParseTimeExactMatch(layout, value string) (time.Time, error) {
-	ts, err := time.Parse(layout, value)
-	if err != nil {
-		return time.Time{}, err
-	}
+func ParseExtendedDateTime(val string, kindType ExtendedTimeKindType) (*ExtendedTime, error) {
```
nathan-artie, line 41: `value` for consistency?

nathan-artie (approved):

PR #904, opened 2024-09-12
[Typing] Default Value

Tang8330:
With this PR: https://github.com/artie-labs/transfer/pull/893, we now treat columns that are `NUMERIC(p, 0)` which are integer columns as `ExtendedDecimal` type with scale of 0.

This means that we can have default values that are integers. This PR adds handling for that.

### DIFF
--- a/clients/shared/default_value.go
+++ b/clients/shared/default_value.go
@@ -39,6 +39,11 @@ func DefaultValue(column columns.Column, dialect sql.Dialect) (any, error) {
 			return sql.QuoteLiteral(extTime.String(column.KindDetails.ExtendedTimeDetails.Format)), nil
 		}
 	case typing.EDecimal.Kind:
+		int64Value, err := typing.AssertType[int64](column.DefaultValue())
+		if err == nil {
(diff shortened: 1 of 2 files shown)


clients/shared/default_value.go
```diff
@@ -39,6 +39,13 @@ func DefaultValue(column columns.Column, dialect sql.Dialect) (any, error) {
...
 		}
 	case typing.EDecimal.Kind:
+		if column.KindDetails.ExtendedDecimalDetails.Scale() == 0 {
+			switch column.DefaultValue().(type) {
+			case int, int8, int16, int32, int64:
+				return fmt.Sprint(column.DefaultValue()), nil
+			}
+		}
+
 		decimalValue, err := typing.AssertType[*decimal.Decimal](column.DefaultValue())
```
danafallon, line 49: Would this just error out before if the column did have an int default value?

Tang8330: Yup!

danafallon (approved):

PR #908, opened 2024-09-17
Remove outdated TODOs

Tang8330:
Removing outdated TODOs


lib/typing/ext/variables.go
```diff
@@ -29,7 +29,6 @@ var supportedDateFormats = []string{
 	PostgresDateFormat,
 }
 
-// TODO: Remove
 var SupportedTimeFormatsLegacy = []string{
```
nathan-artie, line 32: Should we rename this to `SupportedTimeFormats` if we're going to keep it?

nathan-artie (approved):

lib/typing/ext/variables.go
Tang8330: Yup sgtm

PR #911, opened 2024-09-17
Adding Ordered Map

Tang8330:
So we can start to use ordered maps and case insensitive maps inside of Transfer.

### DIFF
--- a/lib/maputil/ordered_map.go
+++ b/lib/maputil/ordered_map.go
@@ -1,0 +1,85 @@
+package maputil
+
+import (
+	"iter"
+	"slices"
+	"strings"
+)
+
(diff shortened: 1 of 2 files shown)


lib/maputil/ordered_map.go
```diff
@@ -0,0 +1,85 @@
+package maputil
```
Tang8330, line 1: This is ported over from our dashboard

danafallon (approved):
Are you going to remove the one in dashboard and make it import this instead?

Tang8330:
> Are you going to remove the one in dashboard and make it import this instead?

Yup

PR #915, opened 2024-09-19
Support column hashing

Tang8330:
Support the ability to hash a column


nathan-artie (approved):

Tang8330:
@nathan-artie Can you take another look? I moved where we're hashing

nathan-artie (approved):

PR #918, opened 2024-09-19
Adding an additional timestamp layout.

Tang8330:
After this PR: https://github.com/artie-labs/transfer/pull/896, we are now strict about time parsing.

As such, this PR adds support for [UTC layout](https://www.utctime.net/z-time-now)

### DIFF
--- a/lib/typing/ext/variables.go
+++ b/lib/typing/ext/variables.go
@@ -11,8 +11,16 @@ const (
 )
 
 var supportedDateTimeLayouts = []string{
+	// UTC
 	RFC3339MillisecondUTC,
+	RFC3339MicrosecondUTC,
+	RFC3339NanosecondUTC,
+	// RFC 3339
+	RFC3339Millisecond,
+	RFC3339Microsecond,
+	RFC3339Nanosecond,
 	time.RFC3339Nano,
+	// Others
 	ISO8601,
 	time.Layout,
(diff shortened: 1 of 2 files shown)


lib/typing/ext/variables.go
```diff
@@ -37,7 +38,8 @@ var SupportedTimeFormats = []string{
 
 // RFC3339 variants
 const (
-	RFC3339Millisecond = "2006-01-02T15:04:05.000Z07:00"
-	RFC3339Microsecond = "2006-01-02T15:04:05.000000Z07:00"
-	RFC3339Nanosecond  = "2006-01-02T15:04:05.000000000Z07:00"
+	RFC3339MillisecondUTC = "2006-01-02T15:04:05.000Z"
+	RFC3339Millisecond    = "2006-01-02T15:04:05.000Z07:00"
+	RFC3339Microsecond    = "2006-01-02T15:04:05.000000Z07:00"
```
nathan-artie, line 43: Should we also add `RFC3339MicrosecondUTC` and `RFC3339Nanosecond UTC`?

nathan-artie (approved):

lib/typing/ext/variables.go
Tang8330: Doesn't hurt, I'll add!

PR #928, opened 2024-09-25
[Redshift] Support truncating values

Tang8330:
This PR adds the ability to truncate exceeded values instead of returning `__artie_exceeded_value`.

This capability could be used for other destinations, however this PR adds it for just Redshift to start.


clients/redshift/cast_test.go
```diff
@@ -13,52 +13,104 @@ import (
...
+					Kind:                    typing.String.Kind,
+					OptionalStringPrecision: typing.ToPtr(int32(3)),
+				}
+
+				assert.Equal(r.T(), "hel", replaceExceededValues("hello", stringKd, true))
+			}
+		}
+		{
+			// Struct and masked
+			assert.Equal(r.T(), fmt.Sprintf(`{"key":"%s"}`, constants.ExceededValueMarker), replaceExceededValues(fmt.Sprintf(`{"foo": "%s"}`, stringutil.Random(int(maxRedshiftLength)+1)), typing.Struct, false))
```
danafallon, line 62: Should we add a test case where a struct value gets truncated?

danafallon (approved):

clients/redshift/cast_test.go
Tang8330: > Should we add a test case where a struct value gets truncated?

They don't get truncated right now because it's unclear how we would actually truncate the struct (e.g. which key do we truncate and still maintain valid JSON?)

PR #930, opened 2024-09-26
Emit when the CDC event was executed

Tang8330:
Emit this metric to Datadog, so we can see if there's a lag in reading source data.


models/event/event.go
```diff
@@ -86,16 +88,30 @@ func ToMemoryEvent(event cdc.Event, pkMap map[string]any, tc kafkalib.TopicConfi
...
 		Data:           hashData(evtData, tc),
 		Deleted:        event.DeletePayload(),
-	}, nil
+	}
+
+	return _event, nil
+}
+
+// EmitExecutionTimeLag - This will check against the current time and the event execution time and emit the lag.
+func (e *Event) EmitExecutionTimeLag(metricsClient base.Client, mode config.Mode) {
```
danafallon, line 106: Does the `mode` need to be passed in, or can we use `e.mode`?

danafallon (approved):

models/event/event.go
Tang8330: Good flag. We should be able to just use `e.mode`!

PR #932, opened 2024-09-26
[Redshift] - Increasing string precision scaffold

Tang8330:
This PR adds the scaffold necessary for us to support increasing string precision for Redshift


Tang8330:
Putting this on draft for a bit

PR #938, opened 2024-09-30
Supporting Databricks - Part one

Tang8330:
## Background

The goal of these PRs is to add Databricks as a supported destination. The main working PR can be found here https://github.com/artie-labs/transfer/pull/937

## Scope

This PR adds the basic Databricks config and tests.

### DIFF
--- a/lib/db/db.go
+++ b/lib/db/db.go
@@ -1,6 +1,7 @@
 package db
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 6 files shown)


lib/db/db.go
```diff
@@ -25,6 +27,28 @@ type storeWrapper struct {
 	*sql.DB
 }
 
+func (s *storeWrapper) ExecContext(ctx context.Context, query string, args ...any) (sql.Result, error) {
```
Tang8330, line 30: We'll need this later for [Databricks](https://docs.databricks.com/en/dev-tools/go-sql-driver.html#manage-files-in-unity-catalog-volumes:~:text=NewContextWithStagingInfo).

nathan-artie (approved):

PR #941, opened 2024-10-02
Cleaning up Dialects

Tang8330:
* Returning an error if we return `typing.Invalid`
* Moved all typing related logic to `dialect/typing` for Snowflake
* Consolidated Snowflake dialect typing tests


Tang8330:
Leaving this for now

PR #946, opened 2024-10-03
github: run tests just once

nathan-artie:
`make race` already runs tests, so we don't need to run tests a second time.

### DIFF
--- a/.github/workflows/gha-go-test.yml
+++ b/.github/workflows/gha-go-test.yml
@@ -20,9 +20,7 @@ jobs:
           make static
       - name: Run vet
         run: make vet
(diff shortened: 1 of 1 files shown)


.github/workflows/gha-go-test.yml
```diff
@@ -12,17 +12,15 @@ jobs:
           go-version: 1.23
       - name: Run staticcheck
         env:
-          SC_VERSION: "2024.1"
+          SC_VERSION: "2024.1.1"
```
nathan-artie, line 15: Also bump `staticcheck`.

danafallon (approved):

PR #954, opened 2024-10-08
[Databricks] Implement sweeping volumes

Tang8330:
Volumes are where we are storing the delta files in Databricks, similar to temporary tables, we should purge them regularly. This PR adds that capability.

### DIFF
--- a/lib/sql/rows.go
+++ b/lib/sql/rows.go
@@ -1,6 +1,9 @@
 package sql
 
-import "database/sql"
+import (
+	"database/sql"
+	"fmt"
+)
 
 func RowsToObjects(rows *sql.Rows) ([]map[string]any, error) {
 	defer rows.Close()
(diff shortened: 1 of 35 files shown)


lib/sql/rows.go
```diff
@@ -0,0 +1,34 @@
...
+func RowsToObjects(rows *sql.Rows) ([]map[string]any, error) {
+	defer rows.Close()
+
+	columns, err := rows.Columns()
+	if err != nil {
+		return nil, err
+	}
+
+	var objects []map[string]any
+	for rows.Next() {
```
nathan-artie, line 14: You'll need to check if `rows.Err()` is not nil after iterating through `rows.Next()`.

nathan-artie: https://github.com/golang/go/blob/356ba0f06586a833cd8de9c04af0d2adddf95851/src/database/sql/sql.go#L3122

Tang8330: sg

lib/sql/rows.go
```diff
@@ -0,0 +1,41 @@
...
+		row := make([]any, len(columns))
+		rowPointers := make([]interface{}, len(columns))
```
nathan-artie, line 19: `any`.

nathan-artie (approved):

PR #955, opened 2024-10-08
[Redshift] Increase String Precision - Part 2

Tang8330:
Part one - https://github.com/artie-labs/transfer/pull/953

## Background

This is the second PR to enable automatic string precision for Redshift. 

## Scope

* This PR adds the ability to update both staging and temporary table's string precision (if necessary).
* Adds the dialect to increase string precision


clients/redshift/staging.go
```diff
@@ -102,22 +114,35 @@ func (s *Store) loadTemporaryTable(tableData *optimization.TableData, newTableID
...
 			if err != nil {
-				return "", err
+				return "", nil, err
 			}
 
-			// TODO: Do something about result.NewLength
+			if result.NewLength > 0 {
+				_newLength, isOk := columnToNewLengthMap[col.Name()]
+				if !isOk || result.NewLength > _newLength {
+					// Update the new length if it's greater than the current one.
```
danafallon, line 123: If the column wasn't in `columnToNewLengthMap` then we could end up setting the length to something lower than it currently is. That could cause problems right?

lib/typing/columns/columns.go
```diff
@@ -126,6 +127,10 @@ func (c *Columns) UpsertColumn(colName string, arg UpsertColumnArg) {
 			col.backfilled = *arg.Backfilled
 		}
 
+		if arg.StringPrecision != nil {
+			col.KindDetails.OptionalStringPrecision = arg.StringPrecision
```
danafallon, line 131: Since this has access to the existing column metadata, is this a better place to check that the new precision is greater than the previous?

clients/redshift/staging.go
Tang8330: [NewLength](https://github.com/artie-labs/transfer/blob/ee1e09e38bbfcb667f4487972b9d07cd00ad39fb/clients/redshift/cast.go#L15) is only populated IFF:

1. expandStringPrecision is enabled (config)
2. Exceeded the current string precision
3. Less than Redshift's max character precision

Tang8330: Anything else would result in NewLength = 0, which would not hit this block.

lib/typing/columns/columns.go
Tang8330: We already have guardrail in how NewLength is being set, I can add something here as well?

clients/redshift/staging.go
danafallon: Ah I see, I thought this was the only place we were checking that but now I see that this is just to handle the edge case where different precisions come through for different rows, to make sure we use the greater one. Carry on!

lib/typing/columns/columns.go
danafallon: My main concern was addressed on the other comment, but yeah it might be nice to add a backup check here too, just in case.

Tang8330: Yeah I don't see why not. I'll add it

Tang8330: I'm gonna open up another PR to add `error` to the signature for UpsertColumn so there's less diff

danafallon (approved):

PR #960, opened 2024-10-10
[Redshift] Moving data type logic into `typing.go`

Tang8330:
No functional changes, just moving files around.


nathan-artie (approved):

lib/typing/decimal/base.go
```diff
@@ -39,7 +39,7 @@ func (d Details) toKind(maxPrecision int32, exceededKind string) string {
 		return exceededKind
 	}
 
-	return fmt.Sprintf("NUMERIC(%v, %v)", d.precision, d.scale)
+	return fmt.Sprintf("NUMERIC(%d, %d)", d.precision, d.scale)
```
nathan-artie, line 42: Just calling out that this is technically a change.

Tang8330: True, my bad - forgot to call this out

PR #962, opened 2024-10-15
Support TIMESTAMP_NTZ

Tang8330:
### Update dialect

- [x] Snowflake
- [x] BigQuery
- [x] Redshift
- [x] Databricks
- [x] MSSQL

### Backwards compatibility

This PR adds backwards compatibility by having a fallback format. 

If the data type in the destination is created as `TIMESTAMP_TZ` and the values in Transfer are stored as `TIMESTAMP_NTZ`, then we should be returning it with the timezone locale to preserve our existing functionality.


lib/debezium/converters/timestamp.go
```diff
@@ -10,7 +10,7 @@ import (
 type Timestamp struct{}
```
Tang8330, line 10: * Timestamp
* MicroTimestamp
* NanoTimestamp

Do not contain any timezone information.

clients/mssql/dialect/dialect.go
```diff
@@ -56,6 +56,8 @@ func (MSSQLDialect) DataTypeForKind(kindDetails typing.KindDetails, isPk bool) s
 	case typing.ETime.Kind:
 		switch kindDetails.ExtendedTimeDetails.Type {
 		case ext.TimestampTzKindType:
+			return "datetimeoffset"
```
danafallon, line 59: It looks like before, if we received a timestamp with tz/offset info, we discarded the tz and stored it without it. Now we'll start storing the tz. Is this change going to result in ddl changes to the column type of existing columns?

Tang8330: Yes it will. I'll migrate the tables that are in MSSQL from `datetime2` -> `datetimeoffset` once this change is in

lib/typing/ext/variables.go
```diff
@@ -44,12 +44,17 @@ var SupportedTimeFormats = []string{
 	AdditionalTimeFormat,
 }
 
+const TimezoneOffset = "Z07:00"
```
danafallon, line 47: Nit: want to call this `TimezoneOffsetFormat`? So it's clear that it's just being used as a format string and not a hardcoded offset that we're using for anything.

clients/mssql/dialect/dialect.go
Tang8330: Actually, I don't think this would impact existing columns. 

If the existing column is created as `datetime2`, we will then tag the column as `TIMESTAMP_NTZ`.
When we merge our in-memory columns with existing, [existing will take precedence](https://github.com/artie-labs/transfer/blob/3c03abf1e0299e7aa37de0d0b4f01a6b51e804bd/lib/optimization/table_data.go#L297).

It is then on the SDK to discard the tz information, which it should be doing already.

lib/optimization/table_data.go
```diff
@@ -293,8 +293,16 @@ func (t *TableData) MergeColumnsFromDestination(destCols ...columns.Column) erro
...
 				}
 
+				// If the column in the destination is a timestamp_tz and the in-memory column is a timestamp_ntz, we should update the layout to contain timezone locale.
+				if foundColumn.KindDetails.ExtendedTimeDetails.Type == ext.TimestampTzKindType && inMemoryCol.KindDetails.ExtendedTimeDetails.Type == ext.TimestampNTZKindType {
+					if inMemoryCol.KindDetails.ExtendedTimeDetails.Format != "" {
+						inMemoryCol.KindDetails.ExtendedTimeDetails.Format += ext.TimezoneOffset
+					}
+				}
+
 				// Just copy over the type since the format wouldn't be present in the destination
```
danafallon, line 303: Is this comment still true?

Tang8330: I think so, we're still not copying the layout from the destination columns. 

The only logic we added is that if the destination column is `TIMESTAMP_TZ` and the in-memory column is `TIMESTAMP_NTZ`, we will append a timezone offset format to the in-memory column's format (if it's set).

clients/mssql/dialect/dialect.go
danafallon: Ah okay so we won't start storing the tz info for existing columns, but if a new column is created that should be tz-aware then we will store it. Seems reasonable 👍

danafallon (approved):

PR #963, opened 2024-10-16
Update UpsertColumn to include an error response

Tang8330:
Adding this in so we can have an extra guardrail to address https://github.com/artie-labs/transfer/pull/955#discussion_r1802191197

### DIFF
--- a/clients/shared/merge.go
+++ b/clients/shared/merge.go
@@ -90,9 +90,14 @@ func Merge(ctx context.Context, dwh destination.DataWarehouse, tableData *optimi
 		for attempts := 0; attempts < backfillMaxRetries; attempts++ {
 			backfillErr = BackfillColumn(dwh, col, tableID)
 			if backfillErr == nil {
-				tableConfig.Columns().UpsertColumn(col.Name(), columns.UpsertColumnArg{
+				err = tableConfig.Columns().UpsertColumn(col.Name(), columns.UpsertColumnArg{
 					Backfilled: typing.ToPtr(true),
 				})
+
+				if err != nil {
+					return fmt.Errorf("failed to update column backfilled status: %w", err)
+				}
+
 				break
 			}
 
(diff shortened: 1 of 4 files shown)


clients/shared/merge.go
```diff
@@ -90,9 +90,14 @@ func Merge(ctx context.Context, dwh destination.DataWarehouse, tableData *optimi
 		for attempts := 0; attempts < backfillMaxRetries; attempts++ {
 			backfillErr = BackfillColumn(dwh, col, tableID)
 			if backfillErr == nil {
-				tableConfig.Columns().UpsertColumn(col.Name(), columns.UpsertColumnArg{
+				err = tableConfig.Columns().UpsertColumn(col.Name(), columns.UpsertColumnArg{
 					Backfilled: typing.ToPtr(true),
 				})
+
+				if err != nil {
```
danafallon, line 97: Could inline this

lib/typing/columns/columns_test.go
```diff
@@ -194,33 +194,33 @@ func TestColumns_UpsertColumns(t *testing.T) {
...
 		PrimaryKey: typing.ToPtr(true),
-	})
+	}))
 	aaaCol, _ := cols.GetColumn("aaa")
 	assert.True(t, aaaCol.ToastColumn)
 	assert.True(t, aaaCol.primaryKey)
 	assert.Equal(t, aaaCol.KindDetails, typing.Invalid)
 
 	length := len(cols.columns)
 	for i := 0; i < 500; i++ {
```
danafallon, line 222: How come we're doing this 500 times?

clients/shared/merge.go
danafallon: Although with the multiple lines that could make it hard to read, so I'm also good with leaving it as is.

danafallon (approved):

lib/typing/columns/columns_test.go
Tang8330: Lol, I was testing it to be idempotent. It's not needed though, I'll remove.

PR #969, opened 2024-10-21
[ExtendedTime] More clean up

Tang8330:
* Addressing PR comments
* Adding the concept of default layout
* Removing NestedKind variables (that were seldomly used)

### DIFF
--- a/lib/typing/ext/time.go
+++ b/lib/typing/ext/time.go
@@ -45,13 +45,6 @@ func NewNestedKind(kindType ExtendedTimeKindType, optionalFormat string) (Nested
 	return NestedKind{Type: kindType, Format: cmp.Or(optionalFormat, defaultLayout)}, nil
 }
 
-var (
-	Time = NestedKind{
-		Type:   TimeKindType,
-		Format: PostgresTimeFormat,
-	}
-)
-
 // ExtendedTime is created because Golang's time.Time does not allow us to explicitly cast values as a date, or time
 // and only allows timestamp expressions.
 type ExtendedTime struct {
--- a/lib/typing/ext/time.go
+++ b/lib/typing/ext/time.go
@@ -52,7 +52,7 @@ type ExtendedTime struct {
 	nestedKind NestedKind
 }
 
-// MarshalJSON is a custom JSON marshaller for ExtendedTime. This is only by MongoDB where we are recursively parsing a nested object.
+// MarshalJSON is a custom JSON marshaller for ExtendedTime.
 func (e ExtendedTime) MarshalJSON() ([]byte, error) {
 	return json.Marshal(e.ts)
 }
--- a/lib/typing/ext/time.go
+++ b/lib/typing/ext/time.go
@@ -36,17 +36,16 @@ type NestedKind struct {
 	Format string
 }
 
+func NewNestedKind(kindType ExtendedTimeKindType, optionalFormat string) (NestedKind, error) {
+	defaultLayout, err := kindType.defaultLayout()
+	if err != nil {
+		return NestedKind{}, err
+	}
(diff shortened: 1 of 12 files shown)


lib/typing/ext/time.go
```diff
@@ -51,29 +52,23 @@ type ExtendedTime struct {
...
-		case TimestampTZKindType:
-			originalFormat = TimestampTZ.Format
-		case TimestampNTZKindType:
-			originalFormat = TimestampNTZ.Format
-		case DateKindType:
-			originalFormat = Date.Format
-		case TimeKindType:
-			originalFormat = Time.Format
-		}
+	defaultLayout, err := kindType.defaultLayout()
```
danafallon, line 62: Could we only do this if `originalFormat` is nil, so we don't hit an error if we're not going to use the default anyway?

lib/typing/ext/time.go
```diff
@@ -17,32 +16,34 @@ const (
...
-	TimestampNTZ = NestedKind{
-		Type:   TimestampNTZKindType,
-		Format: RFC3339NanosecondNoTZ,
-	}
-
-	TimestampTZ = NestedKind{
-		Type:   TimestampTZKindType,
-		Format: time.RFC3339Nano,
+func NewNestedKind(kindType ExtendedTimeKindType, optionalFormat string) (NestedKind, error) {
+	defaultLayout, err := kindType.defaultLayout()
```
danafallon, line 40: same here

lib/typing/ext/time.go
Tang8330: We can, I figure it's safer to error if the `kindType` is something we don't expect.

For example, some random caller passing in "foo" as a `ExtendedTimeKindType`, it's probably better to error when this function gets called rather than somewhere later in downstream

lib/typing/ext/time.go
```diff
@@ -51,29 +52,23 @@ type ExtendedTime struct {
 	nestedKind NestedKind
 }
 
+// MarshalJSON is a custom JSON marshaller for ExtendedTime.
 func (e ExtendedTime) MarshalJSON() ([]byte, error) {
-	return json.Marshal(e.String(""))
+	return json.Marshal(e.ts)
```
danafallon, line 57: Is this not taking the NestedKind's type and format into account?

lib/typing/ext/time_test.go
```diff
@@ -15,7 +15,7 @@ func TestExtendedTime_MarshalJSON(t *testing.T) {
 		// Single value
 		bytes, err := json.Marshal(extTime)
 		assert.NoError(t, err)
-		assert.Equal(t, `"2025-09-13T00:00:00.123Z"`, string(bytes))
+		assert.Equal(t, `"2025-09-13T00:00:00.123456Z"`, string(bytes))
```
danafallon, line 18: re: my last comment - shouldn't this continue being formatted in ms since that's the format specified in `extTime`?

lib/typing/ext/time.go
Tang8330: Ah yeah. I'm purposefully moving away from using the ExtendedTime's nested type and format into account.

My goal is to eventually make ExtendedTime just an alias for `time.Time`.

MarshalJSON was created specifically for nested MongoDB objects and this code is consistent with how Mongo marshals Datetime https://github.com/mongodb/mongo-go-driver/blob/c91b204cc72025ea0c8d0845b84816b8c9e88903/bson/primitive.go#L59-L61

Tang8330: Actually, let me get it to look identical to MongoDB's

danafallon (approved):

PR #976, opened 2024-10-22
Use `time.Time` in more places

Tang8330:
As the title states. The goal is to eventually get rid of `ExtendedTime` altogether.

### DIFF
--- a/clients/shared/default_value_test.go
+++ b/clients/shared/default_value_test.go
@@ -25,7 +25,7 @@ var dialects = []sql.Dialect{
 
 func TestColumn_DefaultValue(t *testing.T) {
 	birthday := time.Date(2022, time.September, 6, 3, 19, 24, 942000000, time.UTC)
-	birthdayExtDateTime, err := ext.ParseDateTime(birthday.Format(ext.ISO8601), ext.TimestampTZKindType)
+	birthdayDateTime, err := ext.ParseDateTime(birthday.Format(ext.ISO8601), ext.TimestampTZKindType)
 	assert.NoError(t, err)
 
 	// date
@@ -85,17 +85,17 @@ func TestColumn_DefaultValue(t *testing.T) {
 		},
 		{
 			name:          "date",
-			col:           columns.NewColumnWithDefaultValue("", dateKind, birthdayExtDateTime),
+			col:           columns.NewColumnWithDefaultValue("", dateKind, birthdayDateTime),
 			expectedValue: "'2022-09-06'",
 		},
 		{
 			name:          "time",
-			col:           columns.NewColumnWithDefaultValue("", timeKind, birthdayExtDateTime),
+			col:           columns.NewColumnWithDefaultValue("", timeKind, birthdayDateTime),
 			expectedValue: "'03:19:24.942'",
 		},
 		{
 			name:          "datetime",
-			col:           columns.NewColumnWithDefaultValue("", dateTimeKind, birthdayExtDateTime),
+			col:           columns.NewColumnWithDefaultValue("", dateTimeKind, birthdayDateTime),
 			expectedValue: "'2022-09-06T03:19:24.942Z'",
 		},
 	}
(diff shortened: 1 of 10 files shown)


clients/shared/default_value_test.go
```diff
@@ -25,7 +25,7 @@ var dialects = []sql.Dialect{
 
 func TestColumn_DefaultValue(t *testing.T) {
 	birthday := time.Date(2022, time.September, 6, 3, 19, 24, 942000000, time.UTC)
-	birthdayExtDateTime, err := ext.ParseExtendedDateTime(birthday.Format(ext.ISO8601), ext.TimestampTZKindType)
+	birthdayExtDateTime, err := ext.ParseDateTime(birthday.Format(ext.ISO8601), ext.TimestampTZKindType)
```
danafallon, line 28: Nit: want to rename `birthdayExtDateTime` to `birthdayDateTime`?

lib/typing/ext/parse_test.go
```diff
@@ -18,7 +18,7 @@ func TestParseFromInterface(t *testing.T) {
 		for _, val := range vals {
 			extTime, err := ParseFromInterface(val, TimestampTZKindType)
```
danafallon, line 19: Rename this var too?

lib/typing/ext/parse_test.go
```diff
@@ -18,7 +18,7 @@ func TestParseFromInterface(t *testing.T) {
...
 			assert.NoError(t, err)
-			assert.Equal(t, val, extTime)
+			assert.Equal(t, val.(*ExtendedTime).GetTime(), extTime)
```
danafallon, line 21: On line 13, can we declare `vals` as `[]ExtendedTime` instead of `[]any` so we don't have to cast `val` here?

danafallon: Oh or will that make `ParseFromInterface` not accept it?

lib/typing/ext/parse_test.go
```diff
@@ -40,72 +40,66 @@ func TestParseFromInterface(t *testing.T) {
...
 }
 
 func TestParseFromInterfaceTime(t *testing.T) {
 	now := time.Now()
 	for _, supportedTimeFormat := range SupportedTimeFormats {
-		et, err := ParseFromInterface(now.Format(supportedTimeFormat), TimeKindType)
+		_time, err := ParseFromInterface(now.Format(supportedTimeFormat), TimeKindType)
 		assert.NoError(t, err)
-		assert.Equal(t, TimeKindType, et.GetNestedKind().Type)
 		// Without passing an override format, this should return the same preserved dt format.
```
danafallon, line 73: Is this comment no longer relevant?

lib/typing/ext/parse_test.go
```diff
@@ -40,72 +40,66 @@ func TestParseFromInterface(t *testing.T) {
...
 
 func TestParseFromInterfaceDate(t *testing.T) {
 	now := time.Now()
 	for _, supportedDateFormat := range supportedDateFormats {
-		et, err := ParseFromInterface(now.Format(supportedDateFormat), DateKindType)
+		_time, err := ParseFromInterface(now.Format(supportedDateFormat), DateKindType)
 		assert.NoError(t, err)
-		assert.Equal(t, DateKindType, et.GetNestedKind().Type)
 
 		// Without passing an override format, this should return the same preserved dt format.
```
danafallon, line 84: Same with this one

danafallon (approved):
A few nitpicks, otherwise lg!

lib/typing/ext/parse_test.go
Tang8330: It can accept `ExtendedTime`!

lib/typing/ext/parse_test.go
Tang8330: Yup, we don't need it anymore!

PR #978, opened 2024-10-23
More refactor around extended time

Tang8330:
1. Specifying layout for Date
2. Updating all the places we used to call `ExtendedTime.String()` to go through `GetTime()`

### DIFF
--- a/lib/debezium/converters/date.go
+++ b/lib/debezium/converters/date.go
@@ -10,16 +10,20 @@ import (
 
 type Date struct{}
 
-func (Date) ToKindDetails() typing.KindDetails {
-	return typing.NewExtendedTimeDetails(typing.ETime, ext.DateKindType, "")
+func (Date) layout() string {
+	return ext.PostgresDateFormat
 }
 
-func (Date) Convert(value any) (any, error) {
+func (d Date) ToKindDetails() typing.KindDetails {
+	return typing.NewExtendedTimeDetails(typing.ETime, ext.DateKindType, d.layout())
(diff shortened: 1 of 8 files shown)


lib/debezium/converters/date.go
```diff
@@ -10,16 +10,20 @@ import (
 
 type Date struct{}
 
-func (Date) ToKindDetails() typing.KindDetails {
-	return typing.NewExtendedTimeDetails(typing.ETime, ext.DateKindType, "")
+func (Date) layout() string {
```
Tang8330, line 13: Specifying a layout

lib/debezium/types_test.go
```diff
@@ -284,14 +284,14 @@ func TestField_ParseValue(t *testing.T) {
 			field := Field{Type: Int64, DebeziumType: MicroTime}
 			value, err := field.ParseValue(int64(54720000000))
 			assert.NoError(t, err)
-			assert.Equal(t, "15:12:00.000000", value.(*ext.ExtendedTime).String(""))
+			assert.Equal(t, "15:12:00.000000", value.(*ext.ExtendedTime).GetTime().Format("15:04:05.000000"))
```
danafallon, line 287: Could we use `MicroTime{}.layout()` here instead of hardcoding the layout?

Tang8330: I think we'll run into a cyclical dependency import since it's not in the same package :(

danafallon (approved):

PR #979, opened 2024-10-23
More refactor around `ExtendedTime`

Tang8330:
More changes, including removing the `.String()` method. 

I added comments to every major change in the PR.

### DIFF
--- a/clients/bigquery/converters/converters.go
+++ b/clients/bigquery/converters/converters.go
@@ -4,13 +4,21 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/artie-labs/transfer/lib/typing"
+
 	"github.com/artie-labs/transfer/lib/typing/decimal"
 	"github.com/artie-labs/transfer/lib/typing/ext"
 )
 
-type StringConverter struct{}
+type StringConverter struct {
+	kd typing.KindDetails
+}
 
-func (StringConverter) Convert(value any) (any, error) {
+func NewStringConverter(kd typing.KindDetails) StringConverter {
+	return StringConverter{kd: kd}
+}
+
+func (s StringConverter) Convert(value any) (any, error) {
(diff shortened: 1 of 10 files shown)


lib/typing/ext/time.go
```diff
@@ -81,8 +81,3 @@ func (e *ExtendedTime) GetTime() time.Time {
 func (e *ExtendedTime) GetNestedKind() NestedKind {
 	return e.nestedKind
 }
-
-func (e *ExtendedTime) String(overrideFormat string) string {
```
Tang8330, line 85: No longer used!

clients/snowflake/staging.go
```diff
@@ -119,18 +119,18 @@ func (s *Store) writeTemporaryTableFile(tableData *optimization.TableData, newTa
 	writer.Comma = '\t'
 
 	columns := tableData.ReadOnlyInMemoryCols().ValidColumns()
-	for _, value := range tableData.Rows() {
-		var row []string
+	for _, row := range tableData.Rows() {
```
Tang8330, line 122: Renamed the variable so it's less confusing

clients/bigquery/converters/converters.go
```diff
@@ -19,7 +26,11 @@ func (StringConverter) Convert(value any) (any, error) {
 	case bool:
 		return fmt.Sprint(castedValue), nil
 	case *ext.ExtendedTime:
-		return castedValue.String(""), nil
+		if err := s.kd.EnsureExtendedTimeDetails(); err != nil {
+			return nil, err
+		}
+
+		return castedValue.GetTime().Format(s.kd.ExtendedTimeDetails.Format), nil
```
Tang8330, line 33: Use the column format instead of `.String()` which relies on the value format.

danafallon (approved):

PR #982, opened 2024-10-24
[ExtendedTime] Addressing TODO

Tang8330:
Addresses TODO from: https://github.com/artie-labs/transfer/pull/979 + adding tests.

### DIFF
--- a/lib/optimization/table_data_test.go
+++ b/lib/optimization/table_data_test.go
@@ -6,8 +6,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/artie-labs/transfer/lib/typing/decimal"
-
 	"github.com/stretchr/testify/assert"
 
 	"github.com/artie-labs/transfer/lib/config"
@@ -15,6 +13,7 @@ import (
 	"github.com/artie-labs/transfer/lib/kafkalib"
 	"github.com/artie-labs/transfer/lib/typing"
 	"github.com/artie-labs/transfer/lib/typing/columns"
+	"github.com/artie-labs/transfer/lib/typing/decimal"
 	"github.com/artie-labs/transfer/lib/typing/ext"
 )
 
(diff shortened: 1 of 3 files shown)


lib/optimization/table_data_test.go
```diff
@@ -331,3 +332,61 @@ func TestTableData_InsertRowSoftDelete(t *testing.T) {
...
+
+		col := mergeColumn(columns.NewColumn("foo", typing.String), decimalCol)
+		assert.Equal(t, details, *col.KindDetails.ExtendedDecimalDetails)
+	}
+	{
+		// Time details get copied over
+		{
+			// Testing for backwards compatibility
+			// in-memory column is TimestampNTZ, destination column is TimestampTZ
+			timestampNTZColumn := columns.NewColumn("foo", typing.MustNewExtendedTimeDetails(typing.ETime, ext.TimestampTZKindType, ""))
```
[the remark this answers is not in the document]
Tang8330: Ah yes, thanks!

lib/typing/ext/time.go
```diff
@@ -20,7 +20,7 @@ func (e ExtendedTimeKindType) defaultLayout() (string, error) {
 	case TimestampTZKindType:
 		return time.RFC3339Nano, nil
 	case TimestampNTZKindType:
-		return RFC3339NanosecondNoTZ, nil
+		return RFC3339NoTZ, nil
```
Tang8330, line 23: This is a deliberate change. The difference between the two are that `RFC3339NoTZ` uses `9` for formatting fractional seconds as opposed to `0`.

Tang8330: The difference between `9` is that if there are no fractional seconds, instead of outputting fractional seconds with all zeroes, it'll simply omit it from the output.

danafallon (approved):

PR #987, opened 2024-10-24
[Extended Time] Preserve timezone (if exists)

Tang8330:
Going through the code and cleaning up Extended time, I found a bug where we are not preserving the timezone from the value.

This PR fixes that and also added a test case to prevent regression.

### DIFF
--- a/lib/debezium/converters/time_test.go
+++ b/lib/debezium/converters/time_test.go
@@ -35,7 +35,7 @@ func TestZonedTimestamp_Convert(t *testing.T) {
 			val, err := ZonedTimestamp{}.Convert("2025-09-13T00:00:12Z")
 			assert.NoError(t, err)
 
-			expectedExtTime := ext.NewExtendedTime(time.Date(2025, time.September, 13, 0, 0, 12, 000000000, time.UTC), ext.TimestampTZKindType, ext.RFC3339)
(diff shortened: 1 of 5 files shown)


lib/debezium/converters/time_test.go
```diff
@@ -35,7 +35,7 @@ func TestZonedTimestamp_Convert(t *testing.T) {
 			val, err := ZonedTimestamp{}.Convert("2025-09-13T00:00:12Z")
 			assert.NoError(t, err)
 
-			expectedExtTime := ext.NewExtendedTime(time.Date(2025, time.September, 13, 0, 0, 12, 000000000, time.UTC), ext.TimestampTZKindType, "2006-01-02T15:04:05.999999999Z")
+			expectedExtTime := ext.NewExtendedTime(time.Date(2025, time.September, 13, 0, 0, 12, 000000000, time.UTC), ext.TimestampTZKindType, ext.RFC3339)
```
Tang8330, line 38: Created a variable so we don't need to hardcode.

lib/typing/ext/variables.go
```diff
@@ -11,10 +11,6 @@ const (
 )
 
 var supportedDateTimeLayouts = []string{
-	// UTC
-	RFC3339MillisecondUTC,
```
Tang8330, line 15: These variants are covered by RFC 3339 below.

danafallon (approved):

PR #999, opened 2024-10-31
[Extended Time] Additional follow up

Tang8330:
Adding support for `DATE` and `TIMESTAMP_NTZ` when the column type is a `STRING` in BigQuery

### DIFF
--- a/clients/bigquery/converters/converters.go
+++ b/clients/bigquery/converters/converters.go
@@ -33,7 +33,7 @@ func (s StringConverter) Convert(value any) (any, error) {
 		case typing.TimestampNTZ:
 			return castedValue.Format(ext.RFC3339NoTZ), nil
 		default:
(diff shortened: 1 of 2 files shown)


clients/bigquery/converters/converters.go
```diff
@@ -25,6 +26,15 @@ func (s StringConverter) Convert(value any) (any, error) {
 		return castedValue.String(), nil
 	case bool, int64:
 		return fmt.Sprint(castedValue), nil
+	case time.Time:
```
danafallon, line 29: Want to add "time.Time" to the error message returned in the default case below?

danafallon (approved):

clients/bigquery/converters/converters.go
Tang8330: Let me modify the error message since it's getting long

PR #1001, opened 2024-11-01
Remove Extended Time - The Finale

Tang8330:
Pulled `TIME` out and fully removed the `ExtendedTime` type!


clients/bigquery/converters/converters_test.go
```diff
@@ -50,6 +49,12 @@ func TestStringConverter_Convert(t *testing.T) {
 			assert.NoError(t, err)
 			assert.Equal(t, "2021-01-01", val)
 		}
+		{
+			// Time
+			val, err := NewStringConverter(typing.Time).Convert(time.Date(2021, 1, 1, 9, 10, 11, 123_456_789, time.UTC))
+			assert.NoError(t, err)
+			assert.Equal(t, "09:10:11.123456Z", val)
```
danafallon, line 56: Is it okay that this has a different tz format than it did before? (`Z` instead of `+00`)

Tang8330: Should be fine. This shouldn't actually impact anyone since the usage of "time" is rare and also this would require them to opt out of the "time" data type and move it to a "string"

lib/optimization/table_data_merge_columns_test.go
```diff
@@ -118,32 +117,35 @@ func TestTableData_UpdateInMemoryColumnsFromDestination(t *testing.T) {
...
-		assert.Equal(t, ext.TimeKindType, timeCol.KindDetails.ExtendedTimeDetails.Type)
-
 		{
-			// Update column from string to TimestampTZ
-			assert.NoError(t, tableData.MergeColumnsFromDestination(columns.NewColumn("string_to_timestamp_tz", typing.TimestampTZ)))
-			col, isOk := tableData.inMemoryColumns.GetColumn("string_to_timestamp_tz")
-			assert.True(t, isOk)
-			assert.Equal(t, typing.TimestampTZ, col.KindDetails)
+			// Time stuff
+			for _, extTimeDetailsCol := range []string{"ext_date", "ext_time", "string_to_timestamp_tz"} {
```
danafallon, line 122: Should we remove `ext` from this variable and from the column names?

Tang8330: Yeah, I'll do that as a f/u since the diff for this PR is already pretty big

lib/typing/ext/parse.go
```diff
@@ -24,78 +24,48 @@ func ParseDateFromInterface(val any) (time.Time, error) {
...
 	default:
 		return time.Time{}, fmt.Errorf("unsupported type: %T", convertedVal)
 	}
 }
 
-func ParseTimestampTZFromInterface(val any) (time.Time, error) {
+func ParseTimestampNTZFromInterface(val any) (time.Time, error) {
 	switch convertedVal := val.(type) {
-	case nil:
-		return time.Time{}, fmt.Errorf("val is nil")
```
danafallon, line 52: Why isn't this one handling the nil case anymore?

Tang8330: Not sure, can add back. Handling or not handling yield will both error

lib/typing/parquet.go
```diff
@@ -69,18 +68,12 @@ type Field struct {
...
-	if k.Kind == Date.Kind {
-		stringKind = true
-	}
-
-	if k.ExtendedTimeDetails != nil && k.ExtendedTimeDetails.Type == ext.TimeKindType {
-		stringKind = true
-	}
-
-	if k.Kind == String.Kind || k.Kind == Struct.Kind || stringKind {
+	switch k.Kind {
```
danafallon, line 71: Can this be combined with the `switch k.Kind` below?

Tang8330: Yes it can :)

danafallon (approved):

PR #1004, opened 2024-11-01
Minor improvements

Tang8330:
Added comments to the improvement areas

### DIFF
--- a/lib/optimization/table_data_merge_columns_test.go
+++ b/lib/optimization/table_data_merge_columns_test.go
@@ -108,46 +108,47 @@ func TestTableData_UpdateInMemoryColumnsFromDestination(t *testing.T) {
 	}
 	{
 		// Casting these as STRING so tableColumn via this f(x) will set it correctly.
-		tableDataCols.AddColumn(columns.NewColumn("ext_date", typing.String))
-		tableDataCols.AddColumn(columns.NewColumn("ext_time", typing.String))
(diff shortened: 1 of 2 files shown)


lib/typing/parquet.go
```diff
@@ -154,7 +154,7 @@ func (k *KindDetails) ParquetAnnotation(colName string) (*Field, error) {
 				},
 			},
 		}, nil
+	default:
```
Tang8330, line 157: This should error so we don't silently not write rows out

lib/optimization/table_data_merge_columns_test.go
```diff
@@ -108,46 +108,46 @@ func TestTableData_UpdateInMemoryColumnsFromDestination(t *testing.T) {
...
 		tableDataCols.AddColumn(columns.NewColumn(strCol, typing.String))
-
 		{
-			// Time stuff
-			for _, extTimeDetailsCol := range []string{"ext_date", "ext_time", "string_to_timestamp_tz"} {
-				col, isOk := tableData.inMemoryColumns.GetColumn(extTimeDetailsCol)
-				assert.True(t, isOk, extTimeDetailsCol)
-				assert.Equal(t, typing.String, col.KindDetails, extTimeDetailsCol)
-			}
+			// Testing converting from string to various time data types
```
Tang8330, line 116: Fast follow from https://github.com/artie-labs/transfer/pull/1001

PR #1006, opened 2024-11-03
`KindForDataType` should return error if unable to map data types

Tang8330:
We were being inconsistent in mapping data types that we get from the destination

### DIFF
--- a/clients/bigquery/dialect/dialect_test.go
+++ b/clients/bigquery/dialect/dialect_test.go
@@ -34,12 +34,34 @@ func TestBigQueryDialect_DataTypeForKind(t *testing.T) {
 func TestBigQueryDialect_KindForDataType(t *testing.T) {
 	dialect := BigQueryDialect{}
 	{
+		// Booleans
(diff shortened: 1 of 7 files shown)


clients/bigquery/dialect/dialect_test.go
```diff
@@ -62,69 +84,85 @@ func TestBigQueryDialect_KindForDataType(t *testing.T) {
...
-		"timestamp": typing.TimestampTZ,
-		"time":      typing.Time,
-		"date":      typing.Date,
-		//Invalid
-		"foo":    typing.Invalid,
-		"foofoo": typing.Invalid,
-		"":       typing.Invalid,
+	{
+		// Integers
+		for _, intKind := range []string{"int", "integer", "int64"} {
```
danafallon, line 106: Before we were testing `"inT64"` (with a capital T) - want to change it back to that so we're still checking it's not case-sensitive?

danafallon (approved):

PR #1025, opened 2024-11-07
Address TODO

Tang8330:
Addressing the TODO from #1024, which is to consolidate how we are casting `value any` into a `string`.

### DIFF
--- a/clients/bigquery/converters/converters_test.go
+++ b/clients/bigquery/converters/converters_test.go
@@ -32,10 +32,19 @@ func TestStringConverter_Convert(t *testing.T) {
 		assert.Equal(t, "true", val)
 	}
 	{
-		// int64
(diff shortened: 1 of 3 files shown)


clients/bigquery/converters/converters_test.go
```diff
@@ -81,7 +85,7 @@ func TestStringConverter_Convert(t *testing.T) {
 			// Time
 			val, err := NewStringConverter(typing.Time).Convert(time.Date(2021, 1, 1, 9, 10, 11, 123_456_789, time.UTC))
 			assert.NoError(t, err)
-			assert.Equal(t, "09:10:11.123456Z", val)
+			assert.Equal(t, "09:10:11.123456", val)
```
Tang8330, line 88: This is now consistent with how we emit `TIME` values which is to not include a TZ locale.

nathan-artie (approved):

PR #1026, opened 2024-11-08
[DDL] Cleaning up primary key creation

Tang8330:
Standing up a function for determining whether we should be creating primary keys or not.

### DIFF
--- a/lib/destination/ddl/ddl_temp_test.go
+++ b/lib/destination/ddl/ddl_temp_test.go
@@ -1,6 +1,7 @@
 package ddl_test
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


lib/destination/ddl/ddl_temp_test.go
```diff
@@ -112,3 +116,114 @@ func (d *DDLTestSuite) TestCreateTemporaryTable() {
 		assert.Contains(d.T(), bqQuery, "CREATE TABLE IF NOT EXISTS `db`.`schema`.`tempTableName` (`foo` string,`bar` float64,`select` string) OPTIONS (expiration_timestamp =")
 	}
 }
+
+func (d *DDLTestSuite) Test_DropTemporaryTableCaseSensitive() {
```
Tang8330, line 120: No changes to these tests. I just moved them from `ddl_test.go`

lib/destination/ddl/ddl_test.go
```diff
@@ -1,123 +1,40 @@
-package ddl_test
+package ddl
```
Tang8330, line 1: Note the pkg name change. I moved all the other tests out so that we can have this be `ddl`. That way, we can have `shouldCreatePrimaryKey` remain a private function.

Once this PR merges, I'm going to try to remove the need for us to specify a different package name.

lib/typing/columns/columns.go
```diff
@@ -73,6 +72,11 @@ func (c *Column) ToLowerName() {
 	c.name = strings.ToLower(c.name)
 }
 
+// SetPrimaryKey is used for tests only.
+func (c *Column) SetPrimaryKey(primaryKey bool) {
```
danafallon, line 76: Maybe we could name this `SetPrimaryKeyForTest` so it's extra clear that we shouldn't use it outside of tests?

danafallon (approved):

lib/typing/columns/columns.go
Tang8330: sgtm

PR #1030, opened 2024-11-10
Refactor Distinct Dates

Tang8330:
## Changes

1. It doesn't need to be a function on `TableData`
2. It can be a private function since it's exclusively used by BigQuery
3. Rewrote tests

### DIFF
--- a/clients/bigquery/partition.go
+++ b/clients/bigquery/partition.go
@@ -1,0 +1,31 @@
+package bigquery
+
+import (
+	"fmt"
+
+	"github.com/artie-labs/transfer/lib/typing/ext"
+)
+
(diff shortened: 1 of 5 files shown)


clients/bigquery/partition.go
```diff
@@ -0,0 +1,31 @@
...
+
+		dateMap[_time.Format(ext.PostgresDateFormat)] = true
+	}
+
+	var distinctDates []string
+	for key := range dateMap {
+		distinctDates = append(distinctDates, key)
+	}
+
+	return distinctDates, nil
```
nathan-artie, line 30: You can do `return slices.Collect(maps.Keys(dateMap))`.

nathan-artie (approved):

PR #1032, opened 2024-11-11
[Debezium] Adding Float64 converter

Tang8330:
Float64 converters are used when we're using Debezium's `double.handling.mode=double`.

Debezium's `double.handling.mode=precise` chokes when there's a NOT NULL column coupled with NaN value.


lib/cdc/relational/debezium.go
```diff
@@ -14,11 +14,11 @@ import (
 type Debezium struct{}
 
 func (Debezium) GetEventFromBytes(bytes []byte) (cdc.Event, error) {
-	var event util.SchemaEventPayload
 	if len(bytes) == 0 {
 		return nil, fmt.Errorf("empty message")
 	}
 
+	var event util.SchemaEventPayload
```
Tang8330, line 21: Just moved this since we don't need to create a variable before the check

PR #1040, opened 2024-11-14
[DDL] More refactor around `AlterTableArgs`

Tang8330:
1. Creating tables is all going through `shared.CreateTable`
2. Removed the create table functionality from `AlterTableArgs`


clients/shared/table.go
```diff
@@ -25,6 +25,6 @@ func CreateTable(ctx context.Context, dwh destination.DataWarehouse, tableData *
 	}
 
 	// Update cache with the new columns that we've added.
-	tc.MutateInMemoryColumns(true, constants.Add, tableData.ReadOnlyInMemoryCols().GetColumns()...)
+	tc.MutateInMemoryColumns(false, constants.Add, tableData.ReadOnlyInMemoryCols().GetColumns()...)
```
Tang8330, line 28: This should have been false to begin with. I'm going to put up another PR to remove this flag.

lib/destination/ddl/ddl.go
```diff
@@ -75,7 +75,6 @@ type AlterTableArgs struct {
 	// ContainsOtherOperations - this is sourced from tableData `containOtherOperations`
 	ContainOtherOperations bool
 	TableID                sql.TableIdentifier
-	CreateTable            bool
```
Tang8330, line 78: I removed the ability to create table in this function

clients/shared/table.go
Tang8330: https://github.com/artie-labs/transfer/pull/1041

danafallon (approved):

PR #1043, opened 2024-11-14
Additional time layout

Tang8330:


lib/typing/ext/variables.go
```diff
@@ -16,6 +16,7 @@ var supportedDateTimeLayouts = []string{
 	RFC3339Nanosecond,
 	time.RFC3339Nano,
```
danafallon, line 17: Just noticed that this one is in here twice

Tang8330: Nice!

Tang8330: I'm gonna write a test to make sure it's only used once

Tang8330: :+1:

danafallon (approved):

PR #1047, opened 2024-11-15
Remove usage of columns in `table_config`

Tang8330:
`*columns.Columns` are now an implementation detail within `table_config`.

Next PR will be to completely remove the need to use `*columns.Columns`


nathan-artie (approved):

Tang8330:
Pausing this PR for a bit. I'm not a fan of the fact that we're using the mutex on the parent object as opposed to the localized columns object.

PR #1049, opened 2024-11-15
[DDL] Standing up `AlterTableAddColumns`

Tang8330:


clients/shared/table.go
```diff
@@ -28,3 +29,22 @@ func CreateTable(ctx context.Context, dwh destination.DataWarehouse, tableData *
 	tc.MutateInMemoryColumns(constants.Add, tableData.ReadOnlyInMemoryCols().GetColumns()...)
 	return nil
 }
+
+func AlterTableAddColumns(ctx context.Context, dwh destination.DataWarehouse, tc *types.DwhTableConfig, tableID sql.TableIdentifier, columns []columns.Column) error {
+	if len(columns) == 0 {
+		return nil
+	}
+
+	sqlParts, addedCols := ddl.BuildAlterTableAddColumns(dwh.Dialect(), tableID, columns)
```
nathan-artie, line 38: Could filter out columns that you want to skip here so that `BuildAlterTableAddColumns( doesn't have to return two things.

nathan-artie (approved):

clients/shared/table.go
Tang8330: True

PR #1064, opened 2024-11-21
Improve Parquet

Tang8330:
Improving the way we write Parquet files, I've added comments to all the changes directly.

### DIFF
--- a/clients/s3/s3.go
+++ b/clients/s3/s3.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/xitongsys/parquet-go-source/local"
 	"github.com/xitongsys/parquet-go/parquet"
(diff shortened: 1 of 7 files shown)


clients/s3/s3.go
```diff
@@ -49,8 +49,7 @@ func (s *Store) IdentifierFor(topicConfig kafkalib.TopicConfig, table string) sq
 func (s *Store) ObjectPrefix(tableData *optimization.TableData) string {
 	tableID := s.IdentifierFor(tableData.TopicConfig(), tableData.Name())
 	fqTableName := tableID.FullyQualifiedName()
-	yyyyMMDDFormat := tableData.LatestCDCTs.Format(ext.PostgresDateFormat)
-
+	yyyyMMDDFormat := fmt.Sprintf("date=%s", time.Now().Format(ext.PostgresDateFormat))
```
Tang8330, line 52: * This should be the current day, not when the CDC event was generated.
* Adding `date=` so that it's consistent with Hive's partitioning scheme: https://lakefs.io/blog/hive-metastore-it-didnt-age-well/#hives-partitioning-scheme

clients/s3/s3.go
```diff
@@ -89,29 +88,24 @@ func (s *Store) Merge(ctx context.Context, tableData *optimization.TableData) er
 		return fmt.Errorf("failed to create a local parquet file: %w", err)
 	}
 
-	pw, err := writer.NewJSONWriter(schema, fw, 4)
+	pw, err := writer.NewCSVWriter(schema, fw, 4)
```
Tang8330, line 91: Using CSV instead since JSON schema for Parquet is not widely supported

lib/parquetutil/parse_values.go
```diff
@@ -9,17 +9,16 @@ import (
 	"github.com/artie-labs/transfer/lib/array"
 	"github.com/artie-labs/transfer/lib/config/constants"
 	"github.com/artie-labs/transfer/lib/typing"
-	"github.com/artie-labs/transfer/lib/typing/columns"
 	"github.com/artie-labs/transfer/lib/typing/decimal"
 	"github.com/artie-labs/transfer/lib/typing/ext"
 )
 
-func ParseValue(colVal any, colKind columns.Column) (any, error) {
+func ParseValue(colVal any, colKind typing.KindDetails) (any, error) {
```
Tang8330, line 16: Using colKind instead, we don't need the whole column

nathan-artie (approved):

PR #1067, opened 2024-11-26
Reducing usages of `columns`

Tang8330:
Reducing more places of us calling `columns` and standardize more around an array of `column` instead.

### DIFF
--- a/lib/optimization/table_data.go
+++ b/lib/optimization/table_data.go
@@ -85,6 +85,17 @@ func (t *TableData) AddInMemoryCol(column columns.Column) {
 	t.inMemoryColumns.AddColumn(column)
 }
 
+func (t *TableData) GetValidColumns() []columns.Column {
+	var cols []columns.Column
+	for _, col := range t.inMemoryColumns.GetColumns() {
+		if col.IsValid() {
+			cols = append(cols, col)
+		}
+	}
+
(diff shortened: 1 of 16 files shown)


lib/typing/columns/columns.go
```diff
@@ -31,16 +31,12 @@ type Column struct {
 	backfilled bool
 }
 
-func (c *Column) PrimaryKey() bool {
-	return c.primaryKey
+func (c Column) IsValid() bool {
+	return c.KindDetails != typing.Invalid
 }
 
-func (c *Column) ShouldSkip() bool {
```
Tang8330, line 38: Removed this function to standardize on `IsValid`.

2 reasons:

1. We should not need to have `Column` be a ptr, removing it allows us to remove the need to check if it's a ptr.
2. Rest of the logic is the same as `IsValid`

lib/optimization/table_data.go
```diff
@@ -85,6 +85,17 @@ func (t *TableData) AddInMemoryCol(column columns.Column) {
 	t.inMemoryColumns.AddColumn(column)
 }
 
+func (t *TableData) GetValidColumns() []columns.Column {
```
Tang8330, line 88: Same functionality as `columns.ValidColumns()`

PR #1072, opened 2024-12-12
[Debezium] Better support around JSON arrays

Tang8330:
Debezium will emit

```json
{
    "type": "array",
    "items": {
        "type": "string",
        "optional": true,
        "name": "io.debezium.data.Json",
        "version": 1
    },
    "optional": true,
    "field": "attributes"
}
``` 

As the field object if the value contains an array of JSON objects. We were previously not doing anything with this information.

As such, we were mainly returning an array of JSON strings (which is what Debezium emits). This PR adds the ability for us to check the `items.name` and if it's JSON, we'll add an additional step to parse the JSON string into an object.

### DIFF
--- a/lib/debezium/schema.go
+++ b/lib/debezium/schema.go
@@ -66,8 +66,8 @@ type Field struct {
 	FieldName    string                `json:"field"`
 	DebeziumType SupportedDebeziumType `json:"name"`
 	Parameters   map[string]any        `json:"parameters"`
-	// [Items] is only populated if the literal type is an array.
-	Items Item `json:"items"`
+	// [ItemsMetadata] is only populated if the literal type is an array.
+	ItemsMetadata Item `json:"items"`
 }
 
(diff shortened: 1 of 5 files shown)


lib/debezium/schema.go
```diff
@@ -53,13 +53,21 @@ const (
...
+
 type Field struct {
 	Type         FieldType             `json:"type"`
 	Optional     bool                  `json:"optional"`
 	Default      any                   `json:"default"`
 	FieldName    string                `json:"field"`
 	DebeziumType SupportedDebeziumType `json:"name"`
 	Parameters   map[string]any        `json:"parameters"`
+	// [Items] is only populated if the literal type is an array.
+	Items Item `json:"items"`
```
danafallon, line 70: `Items` being plural is a little confusing here, maybe we could call this `ItemsMetadata`?

Tang8330: Sg

danafallon (approved):

PR #1079, opened 2024-12-12
[MSSQL] TOAST

Tang8330:
Same rationale as https://github.com/artie-labs/transfer/pull/1078 and https://github.com/artie-labs/transfer/pull/1077.

Matching TOASTED placeholders using LIKE

### DIFF
--- a/clients/mssql/dialect/dialect.go
+++ b/clients/mssql/dialect/dialect.go
@@ -44,13 +44,9 @@ func (MSSQLDialect) IsTableDoesNotExistErr(err error) bool {
 }
 
 func (md MSSQLDialect) BuildIsNotToastValueExpression(tableAlias constants.TableAlias, column columns.Column) string {
+	toastedValue := "%" + constants.ToastUnavailableValuePlaceholder + "%"
 	colName := sql.QuoteTableAliasColumn(tableAlias, column, md)
-	// Microsoft SQL Server doesn't allow boolean expressions to be in the COALESCE statement.
-	if column.KindDetails == typing.Struct {
-		return fmt.Sprintf("COALESCE(%s, {}) != {'key': '%s'}", colName, constants.ToastUnavailableValuePlaceholder)
-	}
-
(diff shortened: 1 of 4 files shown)


clients/mssql/dialect/dialect.go
```diff
@@ -44,13 +44,9 @@ func (MSSQLDialect) IsTableDoesNotExistErr(err error) bool {
...
 func (md MSSQLDialect) BuildIsNotToastValueExpression(tableAlias constants.TableAlias, column columns.Column) string {
+	toastedValue := "%" + constants.ToastUnavailableValuePlaceholder + "%"
 	colName := sql.QuoteTableAliasColumn(tableAlias, column, md)
-	// Microsoft SQL Server doesn't allow boolean expressions to be in the COALESCE statement.
-	if column.KindDetails == typing.Struct {
-		return fmt.Sprintf("COALESCE(%s, {}) != {'key': '%s'}", colName, constants.ToastUnavailableValuePlaceholder)
-	}
-
-	return fmt.Sprintf("COALESCE(%s, '') != '%s'", colName, constants.ToastUnavailableValuePlaceholder)
+	return fmt.Sprintf("COALESCE(%s, '') NOT LIKE '%s'", colName, toastedValue)
```
danafallon, line 49: Could you just do
```
return fmt.Sprintf("COALESCE(%s, '') NOT LIKE '%%s%'", colName, constants.ToastUnavailableValuePlaceholder)
```
instead of defining `toastedValue` above, or do the extra `%`s mess with `fmt.Sprintf`?

danafallon (approved):

clients/mssql/dialect/dialect.go
danafallon: (I guess that's not really more readable though, so I'm also good with it as-is)

Tang8330: Ah I did it like this because `%` is used to start string interpolation

image

danafallon: Ah nice, that's what I suspected

PR #1084, opened 2024-12-16
[Databricks] Preprocess Arrays before `COPY INTO`

Tang8330:
We need to preprocess arrays before we load it into Databricks.


clients/databricks/store.go
```diff
@@ -120,9 +121,40 @@ func (s Store) PrepareTemporaryTable(ctx context.Context, tableData *optimizatio
...
+			ordinalColumn = fmt.Sprintf(`PARSE_JSON(%s)`, ordinalColumn)
+		}
+
+		ordinalColumns = append(ordinalColumns, ordinalColumn)
+	}
+
 	// Copy file from DBFS -> table via COPY INTO, ref: https://docs.databricks.com/en/sql/language-manual/delta-copy-into.html
 	// We'll need \\\\N here because we need to string escape.
-	copyCommand := fmt.Sprintf(`COPY INTO %s BY POSITION FROM '%s' FILEFORMAT = CSV FORMAT_OPTIONS ('escape' = '"', 'delimiter' = '\t', 'header' = 'false', 'nullValue' = '\\\\N')`, tempTableID.FullyQualifiedName(), file.DBFSFilePath())
+	copyCommand := fmt.Sprintf(`
```
Tang8330, line 137: Ran sql format so it's more legible.

Tang8330: Only change here is `(SELECT %s FROM '%s')`

nathan-artie (approved):

PR #1088, opened 2024-12-18
Additional string converters

Tang8330:
Moving and refactoring Int/Float/Decimal converters + adding unit tests


clients/snowflake/snowflake_test.go
```diff
@@ -211,7 +211,7 @@ func (s *SnowflakeTestSuite) TestExecuteMergeDeletionFlagRemoval() {
 	rowsData := make(map[string]map[string]any)
 	for i := 0; i < 5; i++ {
 		rowsData[fmt.Sprintf("pk-%d", i)] = map[string]any{
-			"id":         fmt.Sprintf("pk-%d", i),
```
Tang8330, line 214: This was pretty funny. We labeled `id` as an integer, but the values were `pk-1`, `pk-2`, etc.

lib/typing/converters/util_test.go
```diff
@@ -0,0 +1,68 @@
+package converters
```
Tang8330, line 1: Moved only

lib/typing/converters/util.go
```diff
@@ -0,0 +1,19 @@
+package converters
```
Tang8330, line 1: Moved only

danafallon (approved):

PR #1098, opened 2025-01-02
[BigQuery] Additional guardrails

Tang8330:
If `BQ_AUDIT_ROWS` is enabled, we will perform an additional check to make sure the staging table has more rows than the number of rows in our in-memory database.

### DIFF
--- a/clients/bigquery/bigquery.go
+++ b/clients/bigquery/bigquery.go
@@ -1,19 +1,20 @@
 package bigquery
 
 import (
-	"cloud.google.com/go/bigquery"
-	"cloud.google.com/go/bigquery/storage/managedwriter"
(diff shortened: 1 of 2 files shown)


clients/bigquery/bigquery.go
```diff
@@ -192,6 +214,10 @@ func (s *Store) putTable(ctx context.Context, bqTableID dialect.TableIdentifier,
 			return fmt.Errorf("failed to append rows: %s", status.String())
 		}
 
+		if rowErrs := resp.GetRowErrors(); len(rowErrs) > 0 {
```
Tang8330, line 217: Added an additional guardrail

nathan-artie (approved):

PR #1101, opened 2025-01-04
Improving our Boolean string converter

Tang8330:
Refactoring our Boolean string converter to handle different permutations of 0, 1, "true", "false"

### DIFF
--- a/lib/typing/converters/string_converter.go
+++ b/lib/typing/converters/string_converter.go
@@ -62,9 +62,10 @@ func (BooleanConverter) Convert(value any) (string, error) {
 		return fmt.Sprint(castedValue), nil
 	default:
 		// First try to cast the value into a string and see if we can parse it
-		if fmt.Sprint(value) == "0" {
+		stringValue := strings.ToLower(fmt.Sprint(value))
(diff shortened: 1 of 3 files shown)


lib/typing/converters/string_converter.go
```diff
@@ -57,12 +57,21 @@ func GetStringConverter(kd typing.KindDetails) (Converter, error) {
...
-	if !isOk {
-		return "", fmt.Errorf("failed to cast colVal as boolean, colVal: '%v', type: %T", value, value)
+	switch castedValue := value.(type) {
+	case bool:
+		return fmt.Sprint(castedValue), nil
+	default:
+		// Try to cast the value into a string and see if we can parse it
+		// If not, then return an error
+		switch strings.ToLower(fmt.Sprint(value)) {
+		case "0", "false":
```
danafallon, line 67: Should we also handle "f" and "t"? (not sure if that can happen)

danafallon (approved):

lib/typing/converters/string_converter.go
Tang8330: Hm, let's add it once we see it?

PR #1108, opened 2025-01-25
[Snowflake] Multi-step merge

Tang8330:
This PR implements multi-step merge ("MSM").

MSM is a way for us to be able to increase our ability to flush at a larger interval by having an intermediary table.

The idea here is that you'll be able to specify `flushCounts`.

Steps:

1. If `flushAttempt = 0`, load the data directly into the MSM table.
2. Else, prepare a staging table and merge that into MSM

(Additional step upon last merge attempt), we'll then trigger a merge from MSM -> Target

### DIFF
--- a/lib/kafkalib/topic_test.go
+++ b/lib/kafkalib/topic_test.go
@@ -150,19 +150,10 @@ func TestMultiStepMergeSettings_Validate(t *testing.T) {
 		}.Validate(), "flush count must be greater than 0")
 	}
 	{
-		// Enable, but flush size or table name is empty
-		assert.ErrorContains(t, MultiStepMergeSettings{
-			Enabled:    true,
-			FlushCount: 1,
-			TableName:  "",
-		}.Validate(), "table name is empty")
-	}
-	{
 		// Valid
 		assert.NoError(t, MultiStepMergeSettings{
 			Enabled:    true,
 			FlushCount: 1,
-			TableName:  "test",
 		}.Validate())
 	}
 }
--- a/lib/kafkalib/topic_test.go
+++ b/lib/kafkalib/topic_test.go
@@ -144,7 +144,7 @@ func TestMultiStepMergeSettings_Validate(t *testing.T) {
 		assert.NoError(t, MultiStepMergeSettings{}.Validate())
 	}
 	{
-		// Enable, but flush size or table name is empty
+		// Enable, but flush count and table name are not set
 		assert.ErrorContains(t, MultiStepMergeSettings{
 			Enabled: true,
 		}.Validate(), "flush count must be greater than 0")
--- a/lib/kafkalib/topic_test.go
+++ b/lib/kafkalib/topic_test.go
@@ -145,9 +145,7 @@ func TestMultiStepMergeSettings_Validate(t *testing.T) {
 	}
 	{
 		// Enable, but flush count is not set
-		assert.ErrorContains(t, MultiStepMergeSettings{
-			Enabled: true,
-		}.Validate(), "flush count must be greater than 0")
+		assert.ErrorContains(t, MultiStepMergeSettings{Enabled: true}.Validate(), "flush count must be greater than 0")
 	}
 	{
 		// Valid
--- a/lib/kafkalib/topic_test.go
+++ b/lib/kafkalib/topic_test.go
@@ -144,7 +144,7 @@ func TestMultiStepMergeSettings_Validate(t *testing.T) {
 		assert.NoError(t, MultiStepMergeSettings{}.Validate())
 	}
 	{
-		// Enable, but flush count and table name are not set
+		// Enable, but flush count is not set
 		assert.ErrorContains(t, MultiStepMergeSettings{
 			Enabled: true,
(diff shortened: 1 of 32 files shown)


lib/optimization/table_data.go
```diff
@@ -15,6 +15,20 @@ import (
...
+	return m.FlushAttempts == 0
+}
+
+func (m MultiStepMergeSettings) LastAttempt() bool {
+	return m.FlushAttempts == m.TotalFlushCount
+}
+
+type MultiStepMergeSettings struct {
+	Enabled         bool
+	FlushAttempts   int
```
danafallon, line 28: I wonder if we should make this private and just expose an `increment()` method so it can never get accidentally reset or decremented?

lib/optimization/table_data.go
```diff
@@ -114,6 +146,16 @@ func NewTableData(inMemoryColumns *columns.Columns, mode config.Mode, primaryKey
 		PartitionsToLastMessage: map[string][]artie.Message{},
 		name:                    name,
 	}
+
+	if multiStepMergeSettings := topicConfig.MultiStepMergeSettings; multiStepMergeSettings != nil {
+		td.multiStepMergeSettings = MultiStepMergeSettings{
+			Enabled:         multiStepMergeSettings.Enabled,
+			FlushAttempts:   0,
```
danafallon, line 153: Seems weird to initialize this, can we let it default? (also if we make it private like in my other comment)

Tang8330: We can let it be default (0), I thought doing it this way would be more clear. But I can change

lib/optimization/table_data.go
Tang8330: Sgtm

lib/kafkalib/topic_test.go
```diff
@@ -137,3 +137,23 @@ func TestTopicConfig_Load_ShouldSkip(t *testing.T) {
...
 	}
 }
+
+func TestMultiStepMergeSettings_Validate(t *testing.T) {
+	{
+		// Not enabled
+		assert.NoError(t, MultiStepMergeSettings{}.Validate())
+	}
+	{
+		// Enable, but flush size or table name is empty
```
danafallon, line 147: Should this comment say "but flush count is zero"?

Tang8330: Updated to be more clear

Tang8330: Ah, I removed the table name from the config. Oops!

danafallon (approved):

lib/optimization/table_data.go
```diff
@@ -15,6 +15,28 @@ import (
 	"github.com/artie-labs/transfer/lib/typing/columns"
 )
 
+func (m MultiStepMergeSettings) FirstAttempt() bool {
+	return m.flushAttempts == 0
+}
+
+func (m MultiStepMergeSettings) LastAttempt() bool {
+	return m.flushAttempts == m.TotalFlushCount
```
nathan-artie, line 23: If `m.TotalFlushCount` is say 2 then we'll do two flushes and then on the third flush we'll merge afterwards, is that correct?

lib/kafkalib/topic_test.go
```diff
@@ -137,3 +137,21 @@ func TestTopicConfig_Load_ShouldSkip(t *testing.T) {
...
+	}
+	{
+		// Enable, but flush count is not set
+		assert.ErrorContains(t, MultiStepMergeSettings{Enabled: true}.Validate(), "flush count must be greater than 0")
+	}
+	{
+		// Valid
+		assert.NoError(t, MultiStepMergeSettings{
+			Enabled:    true,
+			FlushCount: 1,
```
nathan-artie, line 154: Is a flush count of 1 in-practice the same as a flush count of 0?

lib/optimization/table_data.go
Tang8330: If total flush count = 2,

We will do the following:

1. create msm table, where count = 0
2. merge staging -> msm, where count = 1
3. merge staging -> msm, where count = 2
4. on the last attempt, it will then merge msm -> target.

So to answer your question, it will do 3 flushes (prepare table on the first attempt and 2 subsequent flushes).

lib/kafkalib/topic_test.go
Tang8330: No it shouldn't be. Flush count starts at the zero index.

nathan-artie: If it's 1 would we not create a staging table, merge into staging, and then merge staging into target all in one go?

Tang8330: When you say 1, are you referring to how many attempts that translates to?

Tang8330: If it's related to `FlushCount = 1`, that will translate into 2 attempts since we're starting at the zero-th index

nathan-artie: I'm referring to if we've defined in the settings that we want to flush attempt 1.

nathan-artie (approved):

PR #1132, opened 2025-02-05
Refactor Event Primary Keys

Tang8330:
This PR removes the need for us to store `pkMap`, which is storing both the primary key(s) and their value(s), which is separate from event data.

Instead, we can store just the list of primary keys and defer the primary key values to be whatever is in the event data.

### DIFF
--- a/models/event/event.go
+++ b/models/event/event.go
@@ -156,6 +156,9 @@ func (e *Event) Validate() error {
 	return nil
 }
 
+// GetPrimaryKeys is returned in a sorted manner to be safe.
+// We use PrimaryKeyValue() as our internal identifier within our db
+// It is critical to make sure `PrimaryKeyValue()` is a deterministic call.
 func (e *Event) GetPrimaryKeys() []string {
 	slices.Sort(e.primaryKeys)
 	return e.primaryKeys
--- a/models/event/event.go
+++ b/models/event/event.go
@@ -23,17 +23,23 @@ import (
 )
 
 type Event struct {
-	Table         string
-	PrimaryKeyMap map[string]any
-	Data          map[string]any // json serialized column data
+	Table string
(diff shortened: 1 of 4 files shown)


models/event/event.go
```diff
@@ -92,11 +94,17 @@ func ToMemoryEvent(event cdc.Event, pkMap map[string]any, tc kafkalib.TopicConfi
...
+	for key := range pkMap {
+		primaryKeys = append(primaryKeys, key)
+	}
+
 	_event := Event{
-		executionTime:  event.GetExecutionTime(),
-		mode:           cfgMode,
+		executionTime: event.GetExecutionTime(),
+		mode:          cfgMode,
+		primaryKeys:   primaryKeys,
```
nathan-artie, line 105: Can you do `primaryKeys:: slices.Collect(maps.Keys(pkMap))`?

nathan-artie: Could also do `slices.Sorted(maps.Keys(pkMap))` so we only have to sort once.

models/event/event.go
```diff
@@ -144,24 +152,19 @@ func (e *Event) Validate() error {
...
-	var keys []string
-	for key := range e.PrimaryKeyMap {
-		keys = append(keys, key)
-	}
-
-	slices.Sort(keys)
-	return keys
+func (e *Event) GetPrimaryKeys() []string {
+	slices.Sort(e.primaryKeys)
+	return e.primaryKeys
```
nathan-artie, line 160: Could do `return slices.Sorted(slices.Values(e.primaryKeys))` so that we aren't mutating `e`.

nathan-artie (approved):

models/event/event.go
Tang8330: Yep, thanks

PR #1142, opened 2025-02-12
Supporting Iceberg - Part 5

Tang8330:
This PR adds a basic scaffold that implements `Baseline` and additional implementations around our dialect.


lib/apachelivy/client.go
```diff
@@ -178,6 +178,8 @@ func (c Client) waitForSessionToBeReady(ctx context.Context) error {
 		default:
 			return fmt.Errorf("session in unexpected state: %q", resp.State)
 		}
+
+		count++
```
Tang8330, line 182: Missed incrementing this count so we can increase jitter sleep.

nathan-artie (approved):

PR #1144, opened 2025-02-12
Adding a GZIP CSV writer

Tang8330:
Adding a library around creating GZIP CSV files so that we can reuse this for Iceberg.

### DIFF
--- a/clients/redshift/staging.go
+++ b/clients/redshift/staging.go
@@ -82,7 +82,7 @@ func (s *Store) PrepareTemporaryTable(ctx context.Context, tableData *optimizati
 
 func (s *Store) loadTemporaryTable(tableData *optimization.TableData, newTableID sql.TableIdentifier) (string, map[string]int32, error) {
 	filePath := fmt.Sprintf("/tmp/%s.csv.gz", newTableID.FullyQualifiedName())
-	gzipWriter, err := csvwriter.NewFilePath(filePath)
+	gzipWriter, err := csvwriter.NewGzipWriter(filePath)
 	if err != nil {
 		return "", nil, err
 	}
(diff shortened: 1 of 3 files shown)


clients/redshift/staging.go
```diff
@@ -83,17 +82,11 @@ func (s *Store) PrepareTemporaryTable(ctx context.Context, tableData *optimizati
 
 func (s *Store) loadTemporaryTable(tableData *optimization.TableData, newTableID sql.TableIdentifier) (string, map[string]int32, error) {
 	filePath := fmt.Sprintf("/tmp/%s.csv.gz", newTableID.FullyQualifiedName())
-	file, err := os.Create(filePath)
+	gzipWriter, err := csvwriter.NewFilePath(filePath)
 	if err != nil {
 		return "", nil, err
 	}
 
-	defer file.Close()
```
nathan-artie, line 91: Do we need to defer `gzipWrite.Close()` so that if there is an error the file gets closed?

lib/csvwriter/gzip.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (g *GzipWriter) Close() error {
+	g.writer.Flush()
+	if err := g.writer.Error(); err != nil {
+		return err
+	}
+	if err := g.gzip.Close(); err != nil {
+		return err
+	}
+	return g.file.Close()
```
nathan-artie, line 43: If the csv writer or gzip writer fail we still probably want to close the file.

Tang8330: Yeah, I'll add that.

clients/redshift/staging.go
Tang8330: Yeah I think so. I just split this up

nathan-artie (approved):

PR #1152, opened 2025-02-14
[Redshift] Moving COPY into `Dialect`

Tang8330:
This is the first PR that moves the existing copy logic into a function so we can unit test it.

Once this is merged, I'll improve the way we handle null values.

### DIFF
--- a/clients/redshift/dialect/dialect.go
+++ b/clients/redshift/dialect/dialect.go
@@ -228,10 +228,10 @@ func (rd RedshiftDialect) BuildCopyStatement(tableID sql.TableIdentifier, cols [
 		quotedColumns[i] = rd.QuoteIdentifier(col)
 	}
 
-	return fmt.Sprintf(`COPY %s (%s) FROM '%s' DELIMITER '\t' NULL AS '\\N' GZIP FORMAT CSV %s dateformat 'auto' timeformat 'auto';`,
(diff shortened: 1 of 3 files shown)


clients/redshift/dialect/dialect.go
```diff
@@ -221,3 +221,17 @@ JOIN
...
     n.nspname = $1 AND c.relname ILIKE $2 AND c.relkind = 'r';`, []any{schemaName, "%" + constants.ArtiePrefix + "%"}
 }
+
+func (rd RedshiftDialect) BuildCopyStatement(tableID sql.TableIdentifier, cols []string, s3URI string, credentialsClause string) string {
+	quotedColumns := make([]string, len(cols))
+	for i, col := range cols {
+		quotedColumns[i] = rd.QuoteIdentifier(col)
+	}
+
+	return fmt.Sprintf(`COPY %s (%s) FROM '%s' DELIMITER '\t' NULL AS '\\N' GZIP FORMAT CSV %s dateformat 'auto' timeformat 'auto';`,
```
nathan-artie, line 231: Should use `QuoteLiteral(s3URI)` + `%s` instead of `'%s'`.

Tang8330: :+1:

nathan-artie (approved):

PR #1179, opened 2025-03-05
Improving AWS credentials

Tang8330:


clients/iceberg/staging.go
```diff
@@ -35,8 +35,8 @@ func (s Store) uploadToS3(ctx context.Context, fp string) (string, error) {
 	s3URI, err := awslib.UploadLocalFileToS3(ctx, awslib.UploadArgs{
 		Bucket:                     s.config.Iceberg.S3Tables.Bucket,
 		FilePath:                   fp,
-		OverrideAWSAccessKeyID:     &s.config.Iceberg.S3Tables.AwsAccessKeyID,
-		OverrideAWSAccessKeySecret: &s.config.Iceberg.S3Tables.AwsSecretAccessKey,
+		OverrideAWSAccessKeyID:     s.config.Iceberg.S3Tables.AwsAccessKeyID,
```
Tang8330, line 38: Removed the need for a pointer since it doesn't do much

lib/awslib/sts.go
```diff
@@ -0,0 +1,87 @@
...
+	expiresAt          time.Time
+
+	// Arguments to generate the credentials:
+	_awsAccessKeyID     string
+	_awsSecretAccessKey string
+	_awsRoleARN         string
+	_sessionLabel       string
+}
+
+func GenerateSTSCredentials(ctx context.Context, awsAccessKeyID string, awsSecretAccessKey string, roleARN string, sessionLabel string) (Credentials, error) {
```
Tang8330, line 28: Added this so that we can reuse this in other places. The idea is that we'll need to use STS to generate temporary credentials if we're assuming a particular role.

danafallon (approved):

PR #1181, opened 2025-03-06
Allow RS credentials override

Tang8330:
If role arn is passed in, we'll generate temporary credentials and inject it into Redshift's credentials clause.

### DIFF
--- a/clients/redshift/redshift.go
+++ b/clients/redshift/redshift.go
@@ -137,11 +137,16 @@ func LoadRedshift(ctx context.Context, cfg config.Config, _store *db.Store) (*St
 		optionalS3Prefix:  cfg.Redshift.OptionalS3Prefix,
 		configMap:         &types.DestinationTableConfigMap{},
 		config:            cfg,
-
-		Store: store,
+		Store:             store,
 	}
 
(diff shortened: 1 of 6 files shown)


clients/redshift/redshift.go
```diff
@@ -114,13 +131,24 @@ func LoadRedshift(cfg config.Config, _store *db.Store) (*Store, error) {
...
 		optionalS3Prefix:  cfg.Redshift.OptionalS3Prefix,
 		configMap:         &types.DestinationTableConfigMap{},
 		config:            cfg,
 
 		Store: store,
-	}, nil
+	}
+
+	if cfg.Redshift.RoleARN != "" {
+		creds, err := awslib.GenerateSTSCredentials(ctx, os.Getenv("AWS_ACCESS_KEY_ID"), os.Getenv("AWS_SECRET_ACCESS_KEY"), cfg.Redshift.RoleARN, "ArtieTransfer")
```
danafallon, line 145: Should we add some handling for if these env vars are empty?

Tang8330: Yeah we can! I'll add to my PR

danafallon (approved):

PR #1184, opened 2025-03-11
Allow Specify AWS Region

Tang8330:


clients/redshift/redshift.go
```diff
@@ -141,7 +141,7 @@ func LoadRedshift(ctx context.Context, cfg config.Config, _store *db.Store) (*St
 	}
 
 	if cfg.Redshift.RoleARN != "" {
-		for _, requiredEnv := range []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"} {
+		for _, requiredEnv := range []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_REGION"} {
```
danafallon, line 144: Does the env var need to be required? It's not being used in this block

Tang8330: Ah yeah, it's not required in this block. It is a required environment variable given that it's being used by the S3 library. I'll move it up

danafallon (approved):

PR #1199, opened 2025-03-23
[Snowflake] PUT Improvements

Tang8330:


clients/snowflake/snowflake_test.go
```diff
@@ -156,6 +166,14 @@ func (s *SnowflakeTestSuite) TestExecuteMergeReestablishAuth() {
 }
 
 func (s *SnowflakeTestSuite) TestExecuteMerge() {
+	columnNames := []string{
```
Tang8330, line 169: I added this because we were iterating over `colToKindDetailsMap` and ordering was not deterministic

Issue #1054, opened 2024-11-19
panic: runtime error: invalid memory address or nil pointer dereference typing.go:36


kholisrag:
hi, wanna report,
I got below error

```log
Nov 19 05:51:10.500 INF [DDL] Executing query query="CREATE TABLE IF NOT EXISTS `gcp-project-id`.`kholistest_stage_mysql_db_Example1Store`.`instantsell_booking_model_counter___artie_hdddt_1732017070` (`id` int,`date_period` date,`model_id` int,`total_booking` int,`total_booking_summary` int,`is_latest_record` int,`created` datetime,`storage` string,`__artie_delete` bool,`__artie_only_set_delete` bool,PRIMARY KEY (`id`) NOT ENFORCED) OPTIONS (expiration_timestamp = TIMESTAMP(\"2024-11-19 11:51:10 UTC\"))"
Nov 19 05:51:10.587 INF [DDL] Executing query query="CREATE TABLE IF NOT EXISTS `gcp-project-id`.`kholistest_stage_mysql_db_Example1Store`.`inventory_cycle_count_snapshot___artie_1it4q_1732017070` (`id` int,`pending_no_location_refund_count` int,`pending_no_location_non_refund_count` int,`pending_storage_count` int,`created` datetime,`modified` datetime,`time_index` int,`__artie_delete` bool,`__artie_only_set_delete` bool,PRIMARY KEY (`id`) NOT ENFORCED) OPTIONS (expiration_timestamp = TIMESTAMP(\"2024-11-19 11:51:10 UTC\"))"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1bd627e]

goroutine 3434 [running]:
github.com/artie-labs/transfer/clients/bigquery/dialect.BigQueryDialect.DataTypeForKind({}, {{0x27e189a, 0x7}, 0x0, 0x0, 0xc0000f9918}, 0xc0?)
	/Users/robintang/code/go/src/github.com/artie-labs/transfer/clients/bigquery/dialect/typing.go:36 +0x1be
github.com/artie-labs/transfer/lib/destination/ddl.BuildCreateTableSQL({0x2bbe768, 0x4211120}, {0x2b993f8, 0xc000988d20}, 0x1, {0xc000ada990, 0xb}, {0xc000bcf808, 0x38, 0x3f})
[...]

PR #1201, opened 2025-03-24
[BigQuery] Handle extended decimals

Tang8330:
If the target data type is a variable numeric or big numeric, we are returning it as a `EDecimal` without any precision or scale defined. As such, we should be handling the NPE that may arise when `ExtendedDecimalDetails` is nil.

This also fixes https://github.com/artie-labs/transfer/issues/1054


Tang8330:
I'm splitting this PR up into smaller chunks

nathan-artie (approved):

PR #1208, opened 2025-03-25
[Snowflake] Improving PUT & COPY INTO

Tang8330:
## Improvements

Adding improvements around Snowflake PUT & COPY INTO commands:

1. Adding single quotes around the filename for PUT
2. Specifying the file to copy for Snowflake. [1]

## Problem 

[1] - This process was susceptible to silent failures. We noticed that the older versions of Snowflake's SDK could have PUT silently fail but still return a success. We would then issue a COPY INTO from a stage to a table and given that there are no files in the stage, the COPY INTO would then also succeed. This could then lead to us to have inconsistent or missing data if PUT failed.

By specifying the file name as part of COPY INTO, Snowflake will throw an error if the remote file does not exist.
image

### DIFF
--- a/clients/snowflake/staging.go
+++ b/clients/snowflake/staging.go
@@ -140,3 +140,8 @@ func (s *Store) writeTemporaryTableFile(tableData *optimization.TableData, newTa
 	writer.Flush()
 	return File{FilePath: fp, FileName: fileName}, writer.Error()
 }
+
+// BuildPutQuery builds a PUT query to upload a file to Snowflake's staging area
+func (s *Store) BuildPutQuery(filePath string, stageName string) string {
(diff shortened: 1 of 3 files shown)


clients/snowflake/staging.go
```diff
@@ -106,7 +109,7 @@ type File struct {
 }
 
 func (s *Store) writeTemporaryTableFile(tableData *optimization.TableData, newTableID sql.TableIdentifier) (File, error) {
-	fileName := fmt.Sprintf("%s.csv", newTableID.FullyQualifiedName())
+	fileName := fmt.Sprintf("%s.csv", strings.ReplaceAll(newTableID.FullyQualifiedName(), `"`, ""))
```
Tang8330, line 112: Removing unnecessary quotes

danafallon (approved):

PR #1224, opened 2025-04-01
[Databricks] Fixes around COPY INTO

Tang8330:
After running the Databricks integration tests, I've observed that by using the same delta file for COPY INTO in Databricks, it will result in subsequent COPY INTOs having no-ops.

My guess is that there's heavy caching on Databricks end to not process that particular COPY INTO. By using a different name, we're able to bypass this problem.

### DIFF
--- a/clients/databricks/store.go
+++ b/clients/databricks/store.go
@@ -101,13 +101,7 @@ func (s Store) PrepareTemporaryTable(ctx context.Context, tableData *optimizatio
 		return fmt.Errorf("failed to cast temp table ID to TableIdentifier")
 	}
 
-	anotherTempTableID := shared.TempTableID(castedTempTableID)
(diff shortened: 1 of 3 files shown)


clients/databricks/store.go
```diff
@@ -44,8 +44,8 @@ func (s Store) Merge(ctx context.Context, tableData *optimization.TableData) (bo
 	return true, nil
 }
 
-func (s Store) Append(ctx context.Context, tableData *optimization.TableData, useTempTable bool) error {
-	return shared.Append(ctx, s, tableData, types.AdditionalSettings{UseTempTable: useTempTable})
+func (s Store) Append(ctx context.Context, tableData *optimization.TableData, _ bool) error {
```
Tang8330, line 47: `useTempTable` is meant for BigQuery only.

danafallon (approved):

PR #1234, opened 2025-04-04
Adding `RowsAffected` check for `PrepareTemporaryTable`

Tang8330:
We'll now have checks for both appending to an existing table and loading a staging table.


clients/snowflake/staging.go
```diff
@@ -94,6 +95,17 @@ func (s *Store) PrepareTemporaryTable(ctx context.Context, tableData *optimizati
 		return fmt.Errorf("failed to run copy into temporary table: %w", err)
 	}
 
+	fmt.Println("result", result)
```
danafallon, line 98: Remove?

Tang8330: Ah yep

danafallon (approved):

PR #1235, opened 2025-04-04
[Snowflake] Use sqlmock

Tang8330:
counterfeiter for `database/sql` is not good because `sql.Rows` is a struct and not something that we can easily mock.

This PR refactors our Snowflake tests to use `sqlmock` which gives us more control around testing SQL commands. This is a pre-requisite for #1234


clients/snowflake/snowflake_test.go
```diff
@@ -114,33 +105,51 @@ func (s *SnowflakeTestSuite) TestExecuteMergeNilEdgeCase() {
...
 func (s *SnowflakeTestSuite) TestExecuteMergeReestablishAuth() {
-	colToKindDetailsMap := map[string]typing.KindDetails{
-		"id":                                typing.Integer,
-		"name":                              typing.String,
-		constants.DeleteColumnMarker:        typing.Boolean,
-		constants.OnlySetDeleteColumnMarker: typing.Boolean,
-		// Add kindDetails to created_at
-		"created_at": typing.MustParseValue("", nil, time.Now().Format(time.RFC3339Nano)),
-	}
+	colToKindDetailsMap := maputil.NewOrderedMap[typing.KindDetails](true)
```
Tang8330, line 139: Added this to make the columns order deterministic

nathan-artie (approved):

PR #1240, opened 2025-04-08
Using even more `ExecContext`

Tang8330:
This should be the last PR to refactor functions to call `ExecContext` over `Exec`

### DIFF
--- a/lib/destination/ddl/ddl_alter_delete_test.go
+++ b/lib/destination/ddl/ddl_alter_delete_test.go
@@ -191,6 +191,7 @@ func (d *DDLTestSuite) TestAlterDelete_Complete() {
 		allColsMap[allCol] = true
 	}
 
+	assert.True(d.T(), d.fakeSnowflakeStagesStore.ExecContextCallCount() > 0)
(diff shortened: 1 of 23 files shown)


lib/destination/ddl/ddl_alter_delete_test.go
```diff
@@ -191,35 +191,36 @@ func (d *DDLTestSuite) TestAlterDelete_Complete() {
...
 				found = true
 			}
 		}
 
 		assert.True(d.T(), found, execQuery)
 	}
 
-	for i := 0; i < d.fakeBigQueryStore.ExecCallCount(); i++ {
-		execQuery, _ := d.fakeBigQueryStore.ExecArgsForCall(0)
+	for i := 0; i < d.fakeBigQueryStore.ExecContextCallCount(); i++ {
```
Tang8330, line 206: These tests weren't being run because ExecCallContext() was 0, so the loop wasn't running.

Tang8330: So I added an additional assert to make sure length > 0

nathan-artie (approved):

PR #1244, opened 2025-04-09
[Snowflake] Use External Stage

Tang8330:
This PR adds the ability for folks to opt to use Snowflake's external stage as opposed to internal table stages.


clients/snowflake/staging.go
```diff
@@ -132,7 +167,45 @@ type File struct {
 	FileName string
 }
 
+func (s *Store) writeTemporaryTableFileGZIP(tableData *optimization.TableData, newTableID sql.TableIdentifier) (File, error) {
```
Tang8330, line 170: This is the same code as below, but using our gzip writer instead.

nathan-artie (approved):

PR #1245, opened 2025-04-10
Minor refactors

Tang8330:
There are 3 changes here.

1. Introducing a slim data structure called `DatabaseAndSchemaPair` which is similar to `TableID` and using this instead of `kafkalib.TopicConfig`

2. Refactoring `GetUniqueDatabaseAndSchemaPairs` to be more clear

3. Refactoring `Sweep`

### DIFF
--- a/lib/kafkalib/topic.go
+++ b/lib/kafkalib/topic.go
@@ -9,15 +9,19 @@ import (
 	"github.com/artie-labs/transfer/lib/stringutil"
 )
 
-// GetUniqueTopicConfigs - will return a list of unique TopicConfigs based on the database and schema in O(n) time.
-func GetUniqueTopicConfigs(tcs []*TopicConfig) []TopicConfig {
-	var uniqueTopicConfigs []TopicConfig
-	seenMap := make(map[string]bool)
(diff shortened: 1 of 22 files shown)


lib/kafkalib/topic.go
```diff
@@ -9,15 +9,19 @@ import (
...
+	var uniqueTopicConfigs []DatabaseAndSchemaPair
+	seenMap := make(map[DatabaseAndSchemaPair]bool)
 	for _, tc := range tcs {
-		key := fmt.Sprintf("%s###%s", tc.Database, tc.Schema)
+		key := tc.BuildDatabaseAndSchemaPair()
 		if _, isOk := seenMap[key]; !isOk {
-			seenMap[key] = true                                  // Mark this as seen
-			uniqueTopicConfigs = append(uniqueTopicConfigs, *tc) // Now add this to the list
+			seenMap[key] = true
+			uniqueTopicConfigs = append(uniqueTopicConfigs, key)
```
nathan-artie, line 24: If the order does not matter could do `return slices.Collect(maps.Keys(seenMap))`

nathan-artie (approved):

PR #1247, opened 2025-04-11
[BigQuery] Update StorageWriter to be exactly once

Tang8330:
We were previously using the default stream for BigQuery which provides at-least-once semantics and we would then dedupe the rows prior to merging.

However, this hinders our ability to truly audit the number of rows in the staging table and whether they fully made it to the staging table or not.

image

This PR is moving us to the committed stream which will allow us to get exactly-once semantics and an exact rows written count to the staging table, thus providing more reliability and assurances around data consistency.

image

### DIFF
--- a/clients/bigquery/bigquery.go
+++ b/clients/bigquery/bigquery.go
@@ -188,36 +188,6 @@ func (s *Store) GetClient(ctx context.Context) *bigquery.Client {
 	return client
 }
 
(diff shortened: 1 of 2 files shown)


Tang8330:
https://cloud.google.com/bigquery/docs/write-api

nathan-artie (approved):

PR #1255, opened 2025-04-16
Iceberg Fixes

Tang8330:
## Changes

* Iceberg requires identifiers to be lowercased.
* We should be explicit about listing column names during APPEND

### DIFF
--- a/clients/iceberg/dialect/dialect.go
+++ b/clients/iceberg/dialect/dialect.go
@@ -211,5 +211,6 @@ func (IcebergDialect) BuildCreateTemporaryView(viewName string, s3Path string) s
 }
 
 func (id IcebergDialect) BuildAppendToTable(tableID sql.TableIdentifier, viewName string, columns []string) string {
+	// Ref: https://downloads.apache.org/spark/docs/3.1.1/sql-ref-syntax-dml-insert-into.html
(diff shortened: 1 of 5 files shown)


clients/iceberg/dialect/dialect.go
```diff
@@ -210,6 +210,6 @@ func (IcebergDialect) BuildCreateTemporaryView(viewName string, s3Path string) s
 	return fmt.Sprintf("CREATE OR REPLACE TEMPORARY VIEW %s USING csv %s;", viewName, getCSVOptions(s3Path))
 }
 
-func (id IcebergDialect) BuildAppendToTable(tableID sql.TableIdentifier, viewName string) string {
-	return fmt.Sprintf("INSERT INTO %s TABLE %s", tableID.FullyQualifiedName(), viewName)
+func (id IcebergDialect) BuildAppendToTable(tableID sql.TableIdentifier, viewName string, columns []string) string {
+	return fmt.Sprintf("INSERT INTO %s (%s) SELECT %s FROM %s", tableID.FullyQualifiedName(), strings.Join(columns, ", "), strings.Join(columns, ", "), viewName)
```
Tang8330, line 214: Making this explicit, otherwise it will look at the column order of the table and not actually match on the column names.

nathan-artie (approved):

PR #1260, opened 2025-04-17
[Debezium] Better Array Handling

Tang8330:
Array Elements may be nested, so we should be parsing them accordingly.

### DIFF
--- a/lib/debezium/converters/date.go
+++ b/lib/debezium/converters/date.go
@@ -20,5 +20,5 @@ func (d Date) Convert(value any) (any, error) {
 	}
 
 	// Represents the number of days since the epoch.
-	return time.UnixMilli(0).In(time.UTC).AddDate(0, 0, int(valueInt64)), nil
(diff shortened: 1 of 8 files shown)


lib/debezium/converters/date.go
```diff
@@ -20,5 +20,5 @@ func (d Date) Convert(value any) (any, error) {
 	}
 
 	// Represents the number of days since the epoch.
-	return time.UnixMilli(0).In(time.UTC).AddDate(0, 0, int(valueInt64)), nil
+	return time.UnixMilli(0).In(time.UTC).AddDate(0, 0, int(valueInt64)).Format(time.DateOnly), nil
```
Tang8330, line 23: I tweaked this to return the date value instead. This way, we are not relying on `time.Time` object for this value.

Did this because:

1. ParseDateFromAny can handle string values
2. If the data type is `date[]` in PG and we're replicating to BQ, we are storing it as `[]string`.

nathan-artie (approved):

PR #1267, opened 2025-04-21
Adding Column Inclusion

Tang8330:
By specifying column inclusion, this instructs Transfer to only include columns specified as columns to include.

### DIFF
--- a/models/event/event.go
+++ b/models/event/event.go
@@ -50,6 +50,18 @@ func transformData(data map[string]any, tc kafkalib.TopicConfig) map[string]any 
 		delete(data, col)
 	}
 
+	// If column inclusion is specified, then we need to include only the specified columns
+	if len(tc.ColumnsToInclude) > 0 {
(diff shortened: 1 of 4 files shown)


models/event/event.go
```diff
@@ -50,11 +50,47 @@ func transformData(data map[string]any, tc kafkalib.TopicConfig) map[string]any
...
+		}
+
+		return &filteredColumns, nil
+	}
+
+	return cols, nil
+}
+
+func ToMemoryEvent(event cdc.Event, pkMap map[string]any, tc kafkalib.TopicConfig, cfgMode config.Mode) (Event, error) {
+	cols, err := buildFilteredColumns(event, tc)
```
danafallon, line 93: If inclusion & exclusion are being handled here, why does `transformData` need to do it too?

Tang8330: this is only filtering the columns not removing the actual event data

danafallon (approved):

PR #1276, opened 2025-04-24
Refactor `SanitizePayload` to be `ParsePayload`

Tang8330:
We were previously sanitizing incoming JSON values by running this through a JSON parser and then stringifying the results to get rid of duplicate keys.

Given that the value is escaped, if we run `JSON.Marshal()` again, it will try to escape the results again. This is then further exacerbated if the column type is an array and inner fields are JSON. This then leaves us with a Frankenstein value where it's an array of JSON strings (instead of JSON objects).

If we then try to marshal the array, it will lead inner values getting double escaped.

### DIFF
--- a/lib/jsonutil/jsonutil.go
+++ b/lib/jsonutil/jsonutil.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 )
 
-// ParsePayload will take in a JSON string, and return a JSON string that has been sanitized (removed duplicate keys)
 func ParsePayload(val string) (any, error) {
 	// There are edge cases for when this may happen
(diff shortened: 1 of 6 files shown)


lib/jsonutil/jsonutil.go
```diff
@@ -4,8 +4,8 @@ import (
 	"encoding/json"
 )
 
-// SanitizePayload will take in a JSON string, and return a JSON string that has been sanitized (removed duplicate keys)
-func SanitizePayload(val string) (any, error) {
+// ParsePayload will take in a JSON string, and return a JSON string that has been sanitized (removed duplicate keys)
```
danafallon, line 7: Should this be updated since it no longer returns a string?

danafallon (approved):

PR #1279, opened 2025-04-24
Moving `WriteTemporaryTableFile` to be shared

Tang8330:
We are doing a lot of the same boilerplate code for all destinations. This PR starts a shared function that we can extend and reuse for other destinations, starting with Snowflake.

Once this is all linked in here, we can then play around with adding randomness to the file name so that we don't run into a race condition when trying to parallelize our table loading (as the file names may collide with eachother)

### DIFF
--- a/clients/shared/temp_table.go
+++ b/clients/shared/temp_table.go
@@ -2,12 +2,17 @@ package shared
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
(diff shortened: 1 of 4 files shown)


Tang8330:
@nathan-artie Could you take a look at this as well? This is a pre-req for #1280

nathan-artie (approved):

PR #1287, opened 2025-04-25
Disable column default value backfills

Tang8330:
Ability to disable column default value backfill based on env var

### DIFF
--- a/clients/shared/default_value.go
+++ b/clients/shared/default_value.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"os"
+	"strconv"
 	"time"
 
 	bigQueryDialect "github.com/artie-labs/transfer/clients/bigquery/dialect"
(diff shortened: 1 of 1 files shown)


clients/shared/default_value.go
```diff
@@ -74,6 +76,17 @@ func DefaultValue(column columns.Column, dialect sql.Dialect) (any, error) {
 }
 
 func BackfillColumn(ctx context.Context, dest destination.Destination, column columns.Column, tableID sql.TableIdentifier) error {
+	if envVar := os.Getenv("DISABLE_BACKFILL"); envVar != "" {
```
danafallon, line 79: Should we make this env var name more specific so it's clear what it does? Like `DISABLE_DEFAULT_VAL_BACKFILL`

danafallon (approved):

clients/shared/default_value.go
Tang8330: sgtm

PR #1300, opened 2025-04-30
[Redshift] Tweaking Redshift SUPER.

Tang8330:
image

[Source](https://docs.aws.amazon.com/redshift/latest/dg/ingest-super.html#:~:text=Copying%20data%20from%20text%20and%20CSV)

We need to be wrapping literal strings in double quotes when the Redshift data type is SUPER


clients/redshift/cast.go
```diff
@@ -41,7 +41,7 @@ func replaceExceededValues(colVal string, colKind typing.KindDetails, truncateEx
 		// Try again, but use [typing.String] instead.
 		result := replaceExceededValues(colVal, typing.String, truncateExceededValue, expandStringPrecision)
 		if result.Exceeded {
-			result.Value = fmt.Sprintf(`"%s"`, result.Value)
```
Tang8330, line 44: We've got a test for this

danafallon (approved):

PR #1301, opened 2025-04-30
Update Struct

Tang8330:
## Checks

- [x] Iceberg
- [x] Databricks
- [x] Snowflake

Our previous PR: https://github.com/artie-labs/transfer/pull/1276 caused some regression as it was escaping JSON values more upstream. By removing that, we are now putting literal string values into our string converter. 

We had this code originally fix our issue with Redshift, but now starting to see it as well for Snowflake.

### DIFF
--- a/lib/typing/converters/string_converter.go
+++ b/lib/typing/converters/string_converter.go
@@ -22,8 +22,6 @@ type Converter interface {
 type GetStringConverterOpts struct {
 	TimestampTZLayoutOverride  string
 	TimestampNTZLayoutOverride string
-	// Redshift - This is to communicate with the typing.StructConverter so that we can handle double quoting literal string values
-	Redshift bool
 }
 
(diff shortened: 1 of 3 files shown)


lib/typing/converters/string_converter.go
```diff
@@ -46,7 +46,7 @@ func GetStringConverter(kd typing.KindDetails, opts GetStringConverterOpts) (Con
 	case typing.Array.Kind:
 		return ArrayConverter{}, nil
 	case typing.Struct.Kind:
-		return NewStructConverter(opts.Redshift), nil
```
danafallon, line 49: Can we remove `Redshift` from `GetStringConverterOpts` or is it still being used somewhere else?

Tang8330: Yeah I'll remove it. Going to test the rest of the destinations first!

danafallon (approved):

PR #1307, opened 2025-05-05
Standing up a S3 Client

Tang8330:
## Changes

Standing up a S3 Client so that we can use this to upload files to S3 instead of calling `UploadLocalFileToS3`. This will make it easier for us to support more functionality with S3.

With this PR, I started to move Iceberg and S3 to start using this function.

### DIFF
--- a/clients/s3/s3.go
+++ b/clients/s3/s3.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	awsCfg "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/xitongsys/parquet-go-source/local"
 	"github.com/xitongsys/parquet-go/parquet"
 	"github.com/xitongsys/parquet-go/writer"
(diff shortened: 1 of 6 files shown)


clients/s3/s3.go
```diff
@@ -146,12 +138,21 @@ func (s *Store) IsRetryableError(_ error) bool {
 	return false // not supported for S3
 }
 
-func LoadStore(cfg config.Config) (*Store, error) {
-	store := &Store{config: cfg}
+func LoadStore(ctx context.Context, cfg config.Config) (*Store, error) {
+	creds := credentials.NewStaticCredentialsProvider(cfg.S3.AwsAccessKeyID, cfg.S3.AwsSecretAccessKey, "")
```
danafallon, line 142: Will we need to start passing in the session token instead of `""` when we make redshift use this? (looks like that's the only thing that's setting `OverrideAWSSessionToken` in `UploadArgs` for the old function)

Tang8330: Yes we will, but not in the S3 destination file. I'll add another function to build AWS Config given its various ways of authenticating.

danafallon (approved):

PR #1321, opened 2025-05-13
Update default value.

Tang8330:
We need to marshal the incoming default value since we are unmarshaling the JSON value upstream.

I've also added some guardrails to the current default value converter to make it less brittle.

### DIFF
--- a/clients/shared/default_value.go
+++ b/clients/shared/default_value.go
@@ -107,8 +107,6 @@ func BackfillColumn(ctx context.Context, dest destination.Destination, column co
 			return fmt.Errorf("failed to escape default value: %w", err)
 		}
 
-		fmt.Println("defaultVal", defaultVal, "column", column.KindDetails.Kind)
-
 		escapedCol := dialect.QuoteIdentifier(column.Name())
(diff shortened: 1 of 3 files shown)


clients/shared/default_value.go
```diff
@@ -70,9 +76,16 @@ func DefaultValue(column columns.Column, dialect sql.Dialect) (any, error) {
 		return decimalValue.String(), nil
 	case typing.String.Kind:
 		return sql.QuoteLiteral(fmt.Sprint(column.DefaultValue())), nil
+	case
+		typing.Boolean.Kind,
+		typing.Integer.Kind,
+		typing.Float.Kind,
+		typing.EDecimal.Kind:
+		return fmt.Sprint(column.DefaultValue()), nil
+	default:
```
Tang8330, line 85: So we don't get surprised by random unsupported types

lib/debezium/converters/basic.go
```diff
@@ -21,8 +21,6 @@ func (JSON) Convert(value any) (any, error) {
 		return value, nil
 	}
 
-	fmt.Println("valueString", valueString)
```
Tang8330, line 24: Extra

danafallon (approved):

PR #1331, opened 2025-05-21
Include Artie Operation

Tang8330:
If this config is set, we'll include the `__artie_operation` column in the target table.


Tang8330:
I'm going to do general clean up here first

PR #1332, opened 2025-05-21
Standing up `BuildColumnsToKeep`

Tang8330:
This way, we don't have to pass around a bunch of variables and we can just generate the list of columns we want to keep and pass it to `DiffAndFilter`

### DIFF
--- a/lib/typing/columns/diff_test.go
+++ b/lib/typing/columns/diff_test.go
@@ -3,63 +3,59 @@ package columns
 import (
 	"testing"
 
-	"github.com/artie-labs/transfer/lib/config"
-
 	"github.com/artie-labs/transfer/lib/config/constants"
-
 	"github.com/artie-labs/transfer/lib/typing"
-
 	"github.com/stretchr/testify/assert"
 )
 
 func TestShouldSkipColumn(t *testing.T) {
 	{
 		// Test delete column marker with soft delete enabled
-		assert.False(t, shouldSkipColumn(constants.DeleteColumnMarker, true, false, false, config.Mode("")))
+		assert.False(t, shouldSkipColumn(constants.DeleteColumnMarker, []string{constants.DeleteColumnMarker}))
 	}
 	{
 		// Test delete column marker without soft delete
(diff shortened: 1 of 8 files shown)


lib/typing/columns/diff_test.go
```diff
@@ -3,63 +3,59 @@ package columns
...
 	}
 	{
 		// Test only set delete column marker
-		assert.True(t, shouldSkipColumn(constants.OnlySetDeleteColumnMarker, false, false, false, config.Mode("")))
+		assert.True(t, shouldSkipColumn(constants.OnlySetDeleteColumnMarker, []string{}))
 	}
 	{
 		// Test only set delete column marker with soft delete enabled
-		assert.True(t, shouldSkipColumn(constants.OnlySetDeleteColumnMarker, true, false, false, config.Mode("")))
+		assert.True(t, shouldSkipColumn(constants.OnlySetDeleteColumnMarker, []string{}))
```
danafallon, line 26: Should `columnsToKeep` include `constants.DeleteColumnMarker` since soft delete is supposed to be enabled?

lib/typing/columns/diff_test.go
```diff
@@ -3,63 +3,59 @@ package columns
...
 	}
 	{
 		// Test regular column
-		assert.False(t, shouldSkipColumn("firstName", false, false, false, config.Mode("")))
+		assert.False(t, shouldSkipColumn("firstName", []string{}))
 	}
 	{
 		// Test regular column with artie updated at and soft delete enabled
-		assert.False(t, shouldSkipColumn("email", true, true, false, config.Mode("")))
+		assert.False(t, shouldSkipColumn("email", []string{}))
```
danafallon, line 42: `[]string{constants.DeleteColumnMarker, constants.UpdateColumnMarker}`?

Tang8330: Yup, thanks for the flag!

lib/typing/columns/diff_test.go
Tang8330: Yes it should!

PR #1348, opened 2025-05-30
Update Dialect Signature

Tang8330:
This PR is adding the scaffold so that we can build MSM support into Redshift


clients/shared/multi_step_merge.go
```diff
@@ -132,11 +132,6 @@ func merge(ctx context.Context, dwh destination.Destination, tableData *optimiza
 		}
 	}()
 
-	snowflakeDialect, ok := dwh.Dialect().(dialect.SnowflakeDialect)
```
Tang8330, line 135: We already have a check on the parent function

danafallon (approved):

PR #1385, opened 2025-07-21
[Kafka] Adding a guardrail for high watermark check

Tang8330:
If this environment variable is enabled, we'll have an additional guardrail to skip messages where the offset is lower than what we have seen.

This may happen if the Kafka cluster is overloaded and rebalances.

### DIFF
--- a/models/event/event.go
+++ b/models/event/event.go
@@ -13,6 +13,7 @@ import (
 	"github.com/artie-labs/transfer/lib/config"
 	"github.com/artie-labs/transfer/lib/config/constants"
 	"github.com/artie-labs/transfer/lib/cryptography"
+	"github.com/artie-labs/transfer/lib/environ"
 	"github.com/artie-labs/transfer/lib/kafkalib"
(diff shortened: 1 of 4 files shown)


models/event/event.go
```diff
@@ -295,6 +296,20 @@ func (e *Event) Save(cfg config.Config, inMemDB *models.DatabaseData, tc kafkali
...
 	}
 
+	kafkaCheck, err := environ.GetBoolEnv(constants.KafkaHWMEnvVar)
+	if err != nil {
+		return false, "", fmt.Errorf("failed to get kafka check: %w", err)
+	}
+
+	if kafkaCheck {
+		if msg, ok := td.PartitionsToLastMessage[message.Partition()]; ok {
+			if msg.Offset() > message.Offset() {
```
danafallon, line 306: Should we check `>=`?

Tang8330: I don't think so, if the message is the same - then processing is a no-op.

danafallon (approved):

PR #1399, opened 2025-07-24
[postgres] Loading a temporary table

Tang8330:


lib/typing/converters/primitives/converter.go
```diff
@@ -23,6 +24,31 @@ func (Int64Converter) Convert(value any) (int64, error) {
 		return int64(castValue), nil
 	case int64:
 		return castValue, nil
+	case float64:
```
Tang8330, line 27: This portion is from Reader. Once this is merged in, I'll consolidate Reader and Transfer

williamhaw (approved):

PR #1401, opened 2025-07-24
[parquet] Migrate to use Arrow to generate Parquet files

Tang8330:
We are currently using https://github.com/xitongsys/parquet-go, community is lacking and also there's errors like: https://github.com/xitongsys/parquet-go/issues/547 that haven't been fixed for 2 years.

Arrow is much more mature and well maintained by a community: https://github.com/apache/arrow-go


lib/typing/converters/primitives/converter.go
```diff
@@ -54,3 +54,26 @@ func (BooleanConverter) Convert(value any) (bool, error) {
 
 	return false, fmt.Errorf("failed to parse boolean, unsupported type: %T", value)
 }
+
+type Float32Converter struct{}
```
Tang8330, line 58: This is lifted from Reader

PR #1410, opened 2025-07-27
Simplifying Transfer

Tang8330:
Removing unnecessary bloat from the critical path, I've outlined my reasoning for deleting the code with comments in the specific codepaths

### DIFF
--- a/clients/shared/merge.go
+++ b/clients/shared/merge.go
@@ -61,8 +61,6 @@ func Merge(ctx context.Context, dest destination.Destination, tableData *optimiz
 		return fmt.Errorf("failed to drop columns for table %q: %w", tableID.Table(), err)
 	}
 
-	// TODO: Examine whether [AuditColumnsToDelete] still needs to be called.
-	tableConfig.AuditColumnsToDelete(srcKeysMissing)
 	if err = tableData.MergeColumnsFromDestination(tableConfig.GetColumns()...); err != nil {
 		return fmt.Errorf("failed to merge columns from destination: %w for table %q", err, tableData.Name())
 	}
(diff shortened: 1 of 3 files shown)


clients/shared/merge.go
```diff
@@ -61,8 +61,6 @@ func Merge(ctx context.Context, dest destination.Destination, tableData *optimiz
 		return fmt.Errorf("failed to drop columns for table %q: %w", tableID.Table(), err)
 	}
 
-	// TODO: Examine whether [AuditColumnsToDelete] still needs to be called.
-	tableConfig.AuditColumnsToDelete(srcKeysMissing)
```
Tang8330, line 65: This function is not necessary. This function was used such that we would iterate over `columnsToDelete` and see if there are any columns we can expunge from the map.

Column deletion already has safety built-in, so this is unnecessary and complicates the code flow.

The current mechanism for deleting a column (unchanged):

1. If we detect extra column(s) from the destination that does not exist in the source, we will run it through `ShouldDeleteColumn`
2. ShouldDeleteColumn will check the settings on whether we've opted in for columns to be automatically deleted or not. If enabled, it will then add it to `colsToDelete` and with the CDC time.
3. If we see another request after the grace period (CDC time + grace period), then we'll allow the column to be dropped.

lib/destination/types/table_config.go
```diff
@@ -158,9 +121,5 @@ func (d *DestinationTableConfig) AddColumnsToDelete(colName string, ts time.Time
 	d.Lock()
 	defer d.Unlock()
 
-	if d.columnsToDelete == nil {
```
Tang8330, line 161: I removed this as well since this is initialized in `NewDestinationTableConfig`

williamhaw (approved):

PR #1412, opened 2025-07-31
[kafka] Adding an additional guardrail

Tang8330:
There's a bug with https://github.com/segmentio/kafka-go as the reader is not synchronized with committing offset and seeking messages.

```
An example that we have observed is:

1. Fetch message (Kafka reader's internal buffer gets messages from 0-100)
2. Commit offset (100)
3. Reader reading messages async times out
4. Reader's internal buffer rewinds back to offset back to previous (say 90)
5. Reading message succeeds, we then jump to 101
```

There's a risk between steps 3-5 that we may end up processing / skip messages from 90->101 (when fetch succeeds, grabbing the message after the last committed offset)

kafka-go's reader internal queue should be invalidated if the commit succeeded and not ever fall backwards.


williamhaw (approved):

Tang8330:
Closing this PR in favor of https://github.com/artie-labs/transfer/pull/1417

PR #1417, opened 2025-08-08
Better handling on Kafka

Tang8330:
## Problem

We’ve been investigating a rare issue where Kafka rebalances can cause out-of-order delivery for updates to the same row. When this happens, downstream consumers may see stale or missing rows.

This isn’t caused by rebalances themselves (which are normal), but by a bug in our Kafka client library ([kafka-go](https://github.com/segmentio/kafka-go)) that sometimes re-delivers old messages after a rebalance. We’ve confirmed the issue, raised it upstream, and we're pursuing parallel mitigations while we evaluate a long-term switch to a better-maintained client.

### Rebalancing

Rebalances themselves are expected and frequent — they happen anytime a consumer joins or leaves a group.

The real issue is that kafka-go mishandles rebalances:
1. Instead of discarding its internal queue, the client may re-deliver old messages.
2. Once the consumer reconnects, it resumes from the correct offset but has already leaked duplicates to the downstream consumer (Transfer).

What should happen is:
* The SDK should maintain proper delivery guarantees and never re-deliver messages.
* On rebalance, the client should invalidate its queue and re-fetch messages from Kafka.

Because this isn’t happening, Transfer sometimes sees duplicates or stale rows.

* We have filed a bug with the provider: https://github.com/segmentio/kafka-go/issues/1398
* kafka-go is also behind in implementing the rest of KIPs, including supporting static membership support https://github.com/segmentio/kafka-go/issues/939 which would support static membership and not cause any rebalances.
* At the consumer level, we do not have a way to detect rebalancing events. [[KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol)], which has also been flagged and requires the community to guess whether there is a rebalance https://github.com/segmentio/kafka-go/issues/1023

The problem is that this library has been [more or less abandoned since Twilio acquired Segment.](https://github.com/segmentio/kafka-go/issues/1356) Active issues raised by the community are not being addressed and they are falling behind on KIPs.

## Solution

This PR introduces 3 additional guardrails to defend against this issue:

1. We are locking at the consumer level (rather than at .Save()) which is one level higher
2. We've also added a guardrail to keep track of the previous offsets seen and will filter out messages that were redelivered due to a rebalance.
3. At Event.Save() - it will error out if it receives a message that has a lower offset than what was previously processed

### DIFF
--- a/lib/kafkalib/consumer.go
+++ b/lib/kafkalib/consumer.go
@@ -35,19 +35,6 @@ func NewConsumerProviderForTest(consumer Consumer, groupID string) *ConsumerProv
 	}
 }
 
-func (c *ConsumerProvider) LockAndProcess(ctx context.Context, lock bool, do func() error) error {
-	if lock {
-		c.mu.Lock()
-		defer c.mu.Unlock()
-	}
-
-	if err := do(); err != nil {
-		return fmt.Errorf("failed to process: %w", err)
-	}
-
-	return nil
-}
-
 func InjectConsumerProvidersIntoContext(ctx context.Context, cfg *Kafka) (context.Context, error) {
 	kafkaConn := NewConnection(cfg.EnableAWSMSKIAM, cfg.DisableTLS, cfg.Username, cfg.Password, DefaultTimeout)
 	dialer, err := kafkaConn.Dialer(ctx)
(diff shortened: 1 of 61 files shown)


lib/kafkalib/consumer.go
```diff
@@ -2,11 +2,98 @@ package kafkalib
...
+
+	return consumer, nil
+}
+
+func (t *ConsumerProvider) CommitMessage(ctx context.Context, msg kafka.Message) error {
+	return t.Consumer.CommitMessages(ctx, msg)
+}
+
+func (t *ConsumerProvider) FetchMessageAndProcess(ctx context.Context, do func(kafka.Message) error) error {
+	msg, err := t.Consumer.FetchMessage(ctx)
```
Tang8330, line 81: Peek?

williamhaw (approved):

PR #1419, opened 2025-08-09
Refactor flush to be topic specific

Tang8330:
This PR refactors how we are flushing specific tables to be topics based.

For the most part, topic to table is 1:1 and flushing based on specific tables is fine. However, this is not great for fan-in and could cause possible issues downstream.

By moving to be topic specific:

1. Changes nothing for normal 1:1 tables.
2. For fan-in tables, this means that we will flush every table from the unified topic. This is desired because we are committing offsets at the end.

### DIFF
--- a/processes/consumer/flush.go
+++ b/processes/consumer/flush.go
@@ -19,8 +19,8 @@ import (
 type Args struct {
 	// [coolDown] - Is used to skip the flush if the table has been recently flushed.
 	CoolDown *time.Duration
-	// [specificTableID] - Is used to flush a specific table. If this is not set, we'll flush everything.
-	SpecificTableID cdc.TableID
+	// [Topic] - This is the specific topic that you would like to flush.
+	Topic string
 	// [reason] - Is used to track the reason for the flush.
 	Reason string
 }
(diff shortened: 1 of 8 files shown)


processes/consumer/flush.go
```diff
@@ -35,77 +35,79 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
...
 		}
 	}
 
 	// Flush will take everything in memory and call the destination to create temp tables.
 	var wg sync.WaitGroup
-	for tableID, tableData := range allTables {
-		if !args.SpecificTableID.IsEmpty() && tableID != args.SpecificTableID {
+	for topic, tables := range topicToTables {
+		if args.Topic != "" && args.Topic != topic {
 			// If the table is specified within args and the table does not match the database, skip this flush.
```
danafallon, line 52: Want to update this comment?

Tang8330: Yep

processes/consumer/flush.go
```diff
@@ -35,77 +35,79 @@ func Flush(ctx context.Context, inMemDB *models.DatabaseData, dest destination.B
```
danafallon, line 35: There's also a comment above this function that should be updated

PR #1437, opened 2025-08-30
Minor code cleanliness

Tang8330:


lib/cdc/util/relational_event.go
```diff
@@ -112,12 +112,12 @@ func (s *SchemaEventPayload) GetFullTableName() string {
 }
 
 func (s *SchemaEventPayload) GetSourceMetadata() (string, error) {
-	json, err := json.Marshal(s.Payload.Source)
+	val, err := json.Marshal(s.Payload.Source)
```
Tang8330, line 115: Don't use the same value as the pkg

williamhaw (approved):

PR #1438, opened 2025-09-01
Renaming `LatestCDCTs` => `LatestTimestamp` + making it private

Tang8330:
Doing some clean up on `TableData` so that we can prepare for a more efficient storage backend

### DIFF
--- a/lib/cdc/event.go
+++ b/lib/cdc/event.go
@@ -41,10 +41,6 @@ func NewTableID(schema, table string) TableID {
 	}
 }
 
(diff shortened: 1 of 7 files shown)


lib/cdc/event.go
```diff
@@ -41,10 +41,6 @@ func NewTableID(schema, table string) TableID {
 	}
 }
 
-func (t TableID) IsEmpty() bool {
```
Tang8330, line 44: Removed

PR #1443, opened 2025-09-09
[Kafka] More refactor to simplify

Tang8330:
## Context

This PR was created to address 2 fundamental problems:

1. Flush and table data logic was far more complicated than it needed to be
2. Partition information should live at the Consumer Topic level not at the Table Data level since it may cause double processing for tables that are fanned in due to the indeterministic nature of Kafka commit ordering

## Changes

1. Refactoring `partitionToAppliedOffset` to be `map[int]kafka.Message` instead of just storing the individual offset. This allows to use the message to commit the Kafka offset
3. Removed `PartitionsToLastMessage` from `TableData` since it now lives on the Kafka Topic Consumer
4. As part of (2), I've also removed the previous Kafka HWM check since that again lives on the Kafka Topic Consumer now
5. Changed Flush from a goroutine to an errorgroup that allows us to not have to create individual Sentry logs
6. Kafka offset commit is now outside of the table loop (which was problematic for fanned-in tables)

### DIFF
--- a/processes/consumer/flush.go
+++ b/processes/consumer/flush.go
@@ -115,6 +115,7 @@ func FlushSingleTopic(ctx context.Context, inMemDB *models.DatabaseData, dest de
 			}
 
 			for _, table := range tables {
+				slog.Info("Flush success, clearing memory...", slog.String("tableID", table.GetTableID().String()))
 				inMemDB.ClearTableConfig(table.GetTableID())
 			}
 		} else {


models/event/event.go
```diff
@@ -432,16 +417,6 @@ func (e *Event) Save(cfg config.Config, inMemDB *models.DatabaseData, tc kafkali
 	}
 
 	td.InsertRow(pkValueString, e.data, e.deleted)
-
```
Tang8330, line 435: No longer needed

williamhaw (approved):

danafallon (approved):

PR #1449, opened 2025-09-27
Simplify snowflake external stage handling

danafallon:
Just moving some code up so all the handling for external vs internal staging is grouped together

### DIFF
--- a/clients/snowflake/staging.go
+++ b/clients/snowflake/staging.go
@@ -76,6 +76,7 @@ func (s *Store) PrepareTemporaryTable(ctx context.Context, tableData *optimizati
 		}
 	}()
 
+	var tableStageName string
 	if s.useExternalStage() {
 		s3Client, err := s.GetS3Client()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


clients/snowflake/staging.go
```diff
@@ -92,24 +93,21 @@ func (s *Store) PrepareTemporaryTable(ctx context.Context, tableData *optimizati
...
-		// Upload the CSV file to Snowflake internal stage
-		tableStageName := addPrefixToTableName(tempTableID, "%")
-		putQuery := fmt.Sprintf("PUT 'file://%s' @%s", file.FilePath, tableStageName)
-		if _, err = s.ExecContext(ctx, putQuery); err != nil {
-			return fmt.Errorf("failed to run PUT for temporary table: %w", err)
-		}
-	}
 
-	tableStageName := addPrefixToTableName(tempTableID, "%")
-	if s.useExternalStage() {
```
danafallon, line 105: We had two different `if s.useExternalStage()` blocks that don't need to be separated, and the assignment on L104 was redundant to L97

clients/snowflake/staging.go
```diff
@@ -92,24 +93,21 @@ func (s *Store) PrepareTemporaryTable(ctx context.Context, tableData *optimizati
...
 		castedTableID, ok := tempTableID.(dialect.TableIdentifier)
 		if !ok {
 			return fmt.Errorf("failed to cast table identifier: %w", err)
 		}
 
 		// Fix the S3 path by ensuring there's a slash between the stage name and the file name
 		tableStageName = fmt.Sprintf("%s.%s.%s/", castedTableID.Database(), castedTableID.Schema(), filepath.Join(s.config.Snowflake.ExternalStage.Name, s.config.Snowflake.ExternalStage.Prefix))
```
danafallon, line 103: @Tang8330 I think we actually don't want to include the s3 prefix here because it's not part of the stage's name, it's part of its url. Is that right?

Tang8330 (approved):

PR #1475, opened 2025-10-15
[Soft Partitioning] Log warning instead of returning error for future partitions

williamhaw:
> [!NOTE]
> Soft partitioning now logs a warning and writes to future partitions instead of erroring, and the Kafka dialer uses a 30s TCP keepalive.
> 
> - **Soft Partitioning (models/event/event.go)**:
>   - Future partitions (partition time > execution time) now emit a warning via `slog.Warn` and proceed to write to the future partition instead of returning an error.
> - **Kafka Connection (lib/kafkalib/connection.go)**:
>   - Set `Dialer.KeepAlive` to `30 * time.Second` to maintain TCP keepalive.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cf4069708ee64e224396bbfa016db46cfcf994ab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/kafkalib/connection.go
+++ b/lib/kafkalib/connection.go
@@ -64,6 +64,7 @@ func (c Connection) Dialer(ctx context.Context, awsOptFns ...func(options *awsCf
 	dialer := &kafka.Dialer{
 		Timeout:   c.timeout,
 		DualStack: true,
+		KeepAlive: 30 * time.Second,
 	}
 
 	switch c.Mechanism() {


models/event/event.go
```diff
@@ -453,7 +453,12 @@ func BuildSoftPartitionSuffix(
 		if distance == 0 {
 			// Same partition, use base suffix
 		} else if distance < 0 {
-			return "", fmt.Errorf("partition time %v for column %q is in the future of execution time %v", partitionColumnValue, sp.PartitionColumn, executionTime)
+			slog.Warn(
```
Tang8330: Is there a point in flooding our logs if this is expected? Also can we add a future timestamp test case?

